PR #45, opened 2017-06-26
Merge feature/monomer-reg into develop

egao12:
To merge and then create RC/1.12.0 off of develop


src/main/resources/log4j.properties
```diff
@@ -6,7 +6,7 @@ log4j.appender.R.File=../logs/catalina.out
 log4j.rootLogger=INFO, R, stdout
 log4j.appender.R.MaxFileSize=2MB
 log4j.logger.org.springframework=INFO
-log4j.logger.com.labsynch.labseer=INFO
+log4j.logger.com.labsynch.labseer=DEBUG
```
bffrost: Should this be the default setting?

goshiro2010: Default for production deploy should be INFO to reduce impact on performance. It would be great if default for Dev = DEBUG

PR #145, opened 2019-01-05
Fixed issue in SEL with Japanese characters where csv is written prop…

bffrost:
Fixed issue in SEL with Japanese characters where csv is written properly, but then read with system default encoding instead of actual file encoding, resulting in unknown characters and a value_kind ls_type_and_kind foreign key constraint violation.

Tested that both files with and without Japanese characters load successfully with this change.
This also seems to fix the issue with using "µ" in units.


brianbolt:
Going to switch this pull request to `release/1.13.3` for a big fix release of `1.13.3.9`

PR #153, opened 2019-08-02
#581: cmpdreg bulkloader autofail invalid values

brianbolt:


bffrost:
@brianbolt I feel like I'm missing something here. Where is the automatic failure actually applied to the SDF rows containing the invalid values?

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -1076,6 +1082,18 @@ public Parent createParent(CmpdRegMolecule mol, Collection<BulkLoadPropertyMappi
...
 		try{
+			//Currently this section of invalid values passed in will only work for string values
+			for (BulkLoadPropertyMappingDTO mapping : mappings){
+				if(mapping.getInvalidValues() != null) {
+					lookUpProperty = mapping.getDbProperty();
+					lookUpString = getStringValueFromMappings(mol, lookUpProperty, mappings);
+					invalidValues = mapping.getInvalidValues();
+					if(invalidValues.contains(lookUpString)) {
+						logger.error("Found invalid "+lookUpProperty+" with code "+lookUpString+" in list of invalid values passed to service");
+						throw new Exception("Invalid property");
```
bffrost: What information is going to be passed back to the user about this error? Will they just see "11 compounds had the error "Invalid Property" "? Or will they see more information about which invalid value caused this SDF entry to fail?

brianbolt: This should restore the old behavior of ACAS which was to produce an error per invalid SDF entry. Attached are screenshots of an example of an SDF with two compounds that both had the same invalid `Lot Chemist` `"BADNAME"`

Summary:
Screen Shot 2019-08-05 at 4 36 14 PM

Error CSV:
Screen Shot 2019-08-05 at 4 36 34 PM

Summary ZIP files list:
Screen Shot 2019-08-05 at 4 40 15 PM

I also tested the same file with 1 good chemist and 1 bad chemist and the result is that one compound is loaded and ends up in the "registered" summary files and the other compound  errors appropriately and ends up in the "error" summary files. Screenshots:

Summary: 
Screen Shot 2019-08-05 at 4 42 53 PM


Error CSV:
Screen Shot 2019-08-05 at 4 43 32 PM

Registered CSV:
Screen Shot 2019-08-05 at 4 45 49 PM

bffrost: Thanks, I misunderstood which try/catch block this was in. It makes sense now.

PR #161, opened 2020-01-15
fixes #638 set url value when saving value

brianbolt:


brianbolt:
Thats right, this should affect SEL, this is for an uncommonly used route in ACAS.

PR #169, opened 2020-02-28
fixes #656 check for lot corp name dupes when registering

brianbolt:


src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -767,6 +767,20 @@ public Lot validateLot(Lot lot, Collection<BulkLoadPropertyMappingDTO> mappings)
 				}
 			}
 		}
+		//Check if the lot corp name already exists in the database
```
bffrost: How does this differ from the code directly above? It seems like that was intended to address this same case but apparently wasn't working. Do we need both? Or should this code replace the block above between lines 753 and 769?

brianbolt: Thanks, I was confused by this too at first but luckily my test scenario gave me an unexpected result when I added this fix to the lines you are talking about.  Basically the above lines are only checking lot corp names of the lots found when searching by the parent corp name for the new lot.  It's possible that the newly registered lot does not have the same parent id as a previously registered lot corp name. This is a more full proof check.

bffrost: Re-reading it, these are actually checking different things. The one above is checking for two "lot #1"'s but this below is explicitly checking for matching full lot corp name. Most of the time these are equivalent but in some cases the check itself is not the same. We should leave both in.

brianbolt: Yeah, thats what I mean to say.  The scenario I was describing was when I had added the check of corp name to the code above instead of the new lines I eventually added. Thanks for the thorough check.

PR #171, opened 2020-03-06
660 dont update nested lsthings when interaction ignored

brianbolt:


src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -617,7 +617,15 @@ public LsThing updateLsThing(LsThing jsonLsThing) throws StaleObjectStateExcepti
 				if (itxLsThingLsThing.getId() == null){
 					//need to save a new itx
 					if (logger.isDebugEnabled()) logger.debug("saving new itxLsThingLsThing: " + itxLsThingLsThing.toJson());
-					updateNestedSecondLsThing(itxLsThingLsThing);
+
```
bffrost: Interesting - I was expecting a change at line 596, not here, but this might also be a good place to add this.
I meant on the "updateNested**FirstLsThing**" function up on line 596.

brianbolt: Thanks didn't read your comment carefully the first time to get the line number. I just saw the line above.  The change I made is unlikely to be a real scenario -- someone sending in an nested thing with a new and ignored interaction but atleast its covered now.

Just pushed a fix for 596 as well.

brianbolt:
fixes mcneilco/acas#660

PR #175, opened 2020-04-10
fixes #174 remove gvnix

### LINKED ISSUE #174: Remove gvnix libraries
These were added to ACAS for a better UI in the roo admin tool but as the roo admin tool has been used less and less it is unnecessary.  Additionally gvnix menus were not migrated when merging in cmpdreg so they are unused currently.

brianbolt:
killed it!


brianbolt:
Yes, I did check the ROO admin GUI to make sure it still works like it did previous to this commit.  ACAS (experiments, protocols, subject...etc.) work well.  Creg stuff is spotty due to the merging of the code bases from a few months ago.  Still requires some fixing up for all functions but I verified that the one's we had fixed previously that users were still using, continue to work.  Additional fixes to other creg entities require more work but thats a seperate issue.

PR #184, opened 2020-06-03
fixes mcneilco/acas#696 add optional ls kind parameter to generic thing search

brianbolt:
This is an additional commit I made on the generic thing search branch addition from last week.

It adds an optional lsKind parmeter to the generic search route.

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiLsThingController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiLsThingController.java
@@ -69,11 +69,11 @@ public class ApiLsThingController {
 	@RequestMapping(value = "/search", method = RequestMethod.GET)
 	@ResponseBody
 	public ResponseEntity<String> genericLsThingSearch(@RequestParam(value="lsType", required = false) String lsType, @RequestParam("q") String searchQuery,
-														@RequestParam(value = "with", required = false) String with) {
+														@RequestParam(value="lsKind", required = false) String lsKind, @RequestParam(value = "with", required = false) String with) {
 		HttpHeaders headers = new HttpHeaders();
 		headers.add("Content-Type", "application/json");
 		try {
-			Collection<LsThing> results = lsThingService.findLsThingsByGenericMetaDataSearch(searchQuery, lsType);
+			Collection<LsThing> results = lsThingService.findLsThingsByGenericMetaDataSearch(searchQuery, lsType, lsKind);
 			if (with != null) {
 				if (with.equalsIgnoreCase("nestedfull")) {
 					return new ResponseEntity<String>(LsThing.toJsonArrayWithNestedFull(results), headers, HttpStatus.OK);
--- a/src/main/java/com/labsynch/labseer/service/LsThingService.java
+++ b/src/main/java/com/labsynch/labseer/service/LsThingService.java
@@ -62,8 +62,13 @@ public interface LsThingService {
 
 	Collection<LsThing> findLsThingsByGenericMetaDataSearch(String searchQuery);
 
-	Collection<LsThing> findLsThingsByGenericMetaDataSearch(String lsType,
-			String searchQuery);
+	Collection<LsThing> findLsThingsByGenericMetaDataSearch(String searchQuery,
+			String lsType);
+
+	Collection<LsThing> findLsThingsByGenericMetaDataSearch(String searchQuery,
+			String lsType,
+			String lsKind
+	);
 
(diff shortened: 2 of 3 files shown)


src/main/java/com/labsynch/labseer/api/ApiLsThingController.java
```diff
@@ -69,11 +69,11 @@
 	@RequestMapping(value = "/search", method = RequestMethod.GET)
 	@ResponseBody
 	public ResponseEntity<String> genericLsThingSearch(@RequestParam(value="lsType", required = false) String lsType, @RequestParam("q") String searchQuery,
-														@RequestParam(value = "with", required = false) String with) {
+														@RequestParam(value="lsKind", required = false) String lsKind, @RequestParam(value = "with", required = false) String with) {
 		HttpHeaders headers = new HttpHeaders();
 		headers.add("Content-Type", "application/json");
 		try {
-			Collection<LsThing> results = lsThingService.findLsThingsByGenericMetaDataSearch(searchQuery, lsType);
+			Collection<LsThing> results = lsThingService.findLsThingsByGenericMetaDataSearch(searchQuery, lsType, lsKind);
```
bffrost, line 76: If that lsKind parameter is optional, do we need to add any if/else logic? Or is lsKind just passed through as null and handled in LsThingServiceImpl?

src/main/java/com/labsynch/labseer/service/LsThingService.java
```diff
@@ -62,8 +62,13 @@ LsThing saveLsThing(LsThing lsThing, boolean isParent, boolean isBatch,
 
 	Collection<LsThing> findLsThingsByGenericMetaDataSearch(String searchQuery);
 
-	Collection<LsThing> findLsThingsByGenericMetaDataSearch(String lsType,
-			String searchQuery);
+	Collection<LsThing> findLsThingsByGenericMetaDataSearch(String searchQuery,
```
bffrost, line 65: I don't see a need for this lsType only route. Also confused that the argument order has been switched.
Is this being used elsewhere in the code? Do we need to switch argument order there as well?

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -1026,8 +1026,12 @@ public LsThing findParentByBatchEquals(LsThing batch) {
 			String searchQuery) {
 		return findLsThingsByGenericMetaDataSearch(searchQuery, null);
 	}
-	
 
+	@Override
+	public Collection<LsThing> findLsThingsByGenericMetaDataSearch(
+			String searchQuery, String lsType) {
+		return findLsThingsByGenericMetaDataSearch(searchQuery, null, null);
```
bffrost, line 1033: This doesn't make sense. It's taking in a `searchQuery` and `lsType`, then completely ignoring the `lsType`??

src/main/java/com/labsynch/labseer/api/ApiLsThingController.java
brianbolt: Yes the logic is handled further down in the the implementation.  I thought the same thing as you, but it was the pattern that was already their for lsType so I followed it.

src/main/java/com/labsynch/labseer/service/LsThingService.java
brianbolt: Yeah, I also thought it was strange too but it was already there and its only being used in `findLsThingProjectsByGenericMetaDataSearch`.  It works because projects are `lsType:project, lsKind:default`. I suppose the correct fix would be to add an lsKind to that call but I didn't want change too much :
https://github.com/mcneilco/acas-roo-server/blob/1e5dd21e2de0c031c5133863e2aeb93b8a903173/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java#L1039

> Also confused that the argument order has been switched.

I switched it to match the way was actually being used in the one place it was being used.

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
brianbolt: Yeah, thats wrong, i'll fix it.

src/main/java/com/labsynch/labseer/service/LsThingService.java
bffrost: Okay, makes sense.

PR #186, opened 2020-06-12
703 ls thing advanced search by codename

brianbolt:


src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -2364,10 +2391,13 @@ else if (query.getRecordedDateLessThan() != null){
 							Predicate firstThingLabelNotEquals = criteriaBuilder.notEqual(firstThingLabel.<String>get("labelText"), interaction.getThingLabelText());
 							firstItxPredicates.add(firstThingLabelNotEquals);
 						}else if(interaction.getOperator().equals("~")){
-							Predicate firstThingLabelLike = criteriaBuilder.like(firstThingLabel.<String>get("labelText"), interaction.getThingLabelText());
+							Predicate firstThingLabelLike = criteriaBuilder.like(criteriaBuilder.lower(firstThingLabel.<String>get("labelText")), '%' + interaction.getThingLabelText().toLowerCase() + '%');
```
bffrost, line 2394: There's a lot of duplicated code here. Could we combine the "~" and "like" into a single "else if" since they do the same logic? That way the "%" + blah + "%" is only duplicated once instead of twice.

brianbolt: Yeah, thanks, I agree but didn't want to clean it all up here. I note it for the future

PR #192, opened 2020-06-25
190 email from address

brianbolt:
See mcneilco/acas#710 for details.  Seperate pull request coming for the mcneilco/acas repo.

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/AuthorServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/AuthorServiceImpl.java
@@ -334,7 +334,7 @@ public class AuthorServiceImpl implements AuthorService {
 		SimpleMailMessage mail = new SimpleMailMessage();
 		mail.setTo(author.getEmailAddress());
 		mail.setSubject("User Activation");
-		if(propertiesUtilService.getEmailFromaddress() != null) mail.setFrom(propertiesUtilService.getEmailFromaddress());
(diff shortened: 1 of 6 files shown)


src/main/java/com/labsynch/labseer/service/AuthorServiceImpl.java
```diff
@@ -334,6 +334,7 @@ public Author signupAuthor(Author author) {
 		SimpleMailMessage mail = new SimpleMailMessage();
 		mail.setTo(author.getEmailAddress());
 		mail.setSubject("User Activation");
+		if(propertiesUtilService.getEmailFromAddress() != null && !propertiesUtilService.getEmailFromAddress().equals("") ) mail.setFrom(propertiesUtilService.getEmailFromAddress());
```
philippcheung, line 337: If mail.setFrom() is empty, what happens?

philippcheung: Okay, my guess is exception is thrown on Line 345 and we just just get a notice in the log file?  How do ACAS users without back end access find out there's a problem?

brianbolt: java mail defaults to the current user/hostname.  So typically this is ok if you are using gmail because gmail doesn't care about the from address. It uses whoever the authentication user is.  With Amazon SES, the from user is what shows up in the email and the autentication user is just the IAM credentials for the SES user.

brianbolt: Also, if an email fails to send, it shows up in the logs.

PR #194, opened 2020-07-08
713 ignored db properties

brianbolt:


src/main/java/com/labsynch/labseer/dto/BulkLoadPropertyMappingDTO.java
```diff
@@ -29,22 +29,25 @@
 
 	private Collection<String> invalidValues;
 
-    private String defaultVal;
+	private String defaultVal;
+	
+    private boolean ignored;
```
bffrost, line 34: nit: wacky indentation here... Looks like you fixed some on line 32 but it's still there on line 34.

PR #197, opened 2020-10-29
fixes #196 Add error log when unable to parse username from DN.  Add …

### LINKED ISSUE #196: Error when trying to parse first name last name from DN
The default user context mapper for LDAP in ldap-authentication-provider does not give direct access to the CN of the user.  Instead we rely on putting the CN into the DN of the user.  This makes ACAS difficult to work with on many LDAP setups.  This change allows the admin to configure the applicationContext-security.xml file to map users using inetOrgPerson which gives direct access to the CN. 

To configure this, the admin would:

Add the `inetOrgPersonContextMapper` bean into their context security file:

```
  <beans:bean id="postSuccessAuthHandler"
    class="com.labsynch.labseer.service.PostSuccessfulAuthenticationHandler">
  </beans:bean>

  <beans:bean id="inetOrgPersonContextMapper" class="org.springframework.security.ldap.userdetails.InetOrgPersonContextMapper">
  </beans:bean>
```

And add the `user-context-mapper-ref` with value `inetOrgPersonContextMapper ` to their `ldap-authentication-provider` details:
```
    <ldap-authentication-provider server-ref="ldapServer"
      user-dn-pattern=""
      user-search-filter="uid={0}"
      group-search-filter="(|(member={0})(cn={0}))"
      user-search-base="{{ user_search_base }}"
      group-search-base="{{ group_search_base }}"
      user-context-mapper-ref="inetOrgPersonContextMapper"
      />
```

brianbolt:
…inetOrgPerson user class mapper to get easier access to CN when ldap is using inetOrgPerson

### DIFF
--- a/src/main/java/com/labsynch/labseer/utils/SecurityUtils.java
+++ b/src/main/java/com/labsynch/labseer/utils/SecurityUtils.java
@@ -67,7 +67,7 @@ public class SecurityUtils {
 
 	}
 
-	private Author creacreateAuthorteAuthor(Authentication authentication) {
+    private Author createAuthor(Authentication authentication) {
 
 		Object principal = authentication.getPrincipal();
 
--- a/src/main/java/com/labsynch/labseer/utils/SecurityUtils.java
+++ b/src/main/java/com/labsynch/labseer/utils/SecurityUtils.java
@@ -12,6 +12,7 @@ import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.userdetails.User;
 import org.springframework.security.ldap.userdetails.LdapUserDetails;
+import org.springframework.security.ldap.userdetails.InetOrgPerson;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
@@ -66,7 +67,7 @@ public class SecurityUtils {
 
 	}
 
-	private Author createAuthor(Authentication authentication) {
+	private Author creacreateAuthorteAuthor(Authentication authentication) {
 
 		Object principal = authentication.getPrincipal();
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/utils/SecurityUtils.java
```diff
@@ -66,7 +67,7 @@ public void updateAuthorInfo(Authentication authentication) {
 
 	}
 
-	private Author createAuthor(Authentication authentication) {
+	private Author creacreateAuthorteAuthor(Authentication authentication) {
```
bffrost, line 70: copy/paste issue

bffrost (changes requested):

src/main/java/com/labsynch/labseer/utils/SecurityUtils.java
brianbolt: Thanks, somehow did that after building and deploying/testing

src/main/java/com/labsynch/labseer/utils/SecurityUtils.java
```diff
@@ -75,7 +76,30 @@ private Author createAuthor(Authentication authentication) {
 		
 		Author author = new Author();
 
-		if (principal instanceof LdapUserDetails) {
+		if (principal instanceof InetOrgPerson) {
+			InetOrgPerson inetOrgPerson = (InetOrgPerson) principal;
```
bffrost, line 80: Actually it looks like `InetOrgPerson` offers `getSn` and `getGivenName` methods. I think it'd be better if we tried those, then fell back to splitting Cn if we couldn't get those attributes.

brianbolt: Yeah, I checked on that. 

This is our current version:
https://docs.spring.io/spring-security/site/docs/3.2.5.RELEASE/apidocs/org/springframework/security/ldap/userdetails/InetOrgPerson.html

I was just trying to get this fixed for a specific case where the CN wasn't available in the DN.  But I can try updating our spring security to see if that gets us all the way there with this fix.

bffrost (approved):
LGTM. Thanks for doing the digging on spring upgrades and other approaches.

brianbolt:
Thanks!

PR #201, opened 2021-01-30
fixes mcneilco/acas#728 empty mol files should be considered the same…

brianbolt:
… structure


bffrost (approved):
Seems a little hacky but I'm OK with it since it seems to fix a specific use case and I can't think of a broken use case. Theoretically seems like there could be distinct "structures" with query atoms but can't think of an actual use case for that.

brianbolt:
I removed the hackyness by using chemaxons "isEmpty" method and writing a similar method for indigo by checking if a structure has any atoms, bonds or s groups to determine if it is empty or not.  I also changed any use of mol weight being 0.0 to use the new "isEmpty" method

bffrost (approved):
LGTM

PR #206, opened 2021-05-08
741 acas forms blob value support

brianbolt:


src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -1326,7 +1326,7 @@ public void updateLsStates(LsThing jsonLsThing, LsThing updatedLsThing){
 						if (lsThingValue.getLsState() == null) lsThingValue.setLsState(updatedLsThingState);
 						LsThingValue updatedLsThingValue = null;
 						if (lsThingValue.getId() == null){
-							updatedLsThingValue = LsThingValue.create(lsThingValue);
+							updatedLsThingValue = new LsThingValue(lsThingValue);
```
bffrost, line 1329: @brianbolt What's the reasoning behind this change?

brianbolt: Yes.  Should have noted it somewhere in the pull request as it's still a bit of a mystery on why .create is being used in the first place.  LsThingValue.create (link1 below) does this interesting thing where it serializes the value and deserializes it again.  So if you passed in a new blob value, the byte array would get wiped when saving.  I can't speak to why this was there in first place which of course is somewhat dangerous to change without knowing.  I see the "create" method being used elsewhere.  Any ideas?

link1: https://github.com/mcneilco/acas-roo-server/blob/master/src/main/java/com/labsynch/labseer/domain/LsThingValue.java#L222
link2: https://github.com/mcneilco/acas-roo-server/blob/85e0ac690cc690322b8067b239459170aa5f28ae/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java#L1366

bffrost: Thanks. I don't recall specifics, but my best guess is that this was used before the LsThingValue constructor that takes in another LsThingValue was written. https://github.com/mcneilco/acas-roo-server/blob/85e0ac690cc690322b8067b239459170aa5f28ae/src/main/java/com/labsynch/labseer/domain/LsThingValue.java#L54

Going to/from JSON is a shortcut to use the auto-generated JSON parsing classes rather than explicitly listing out all the .get and .set methods as is done in the link I added above.

I did a manual review of the constructor method versus the list of columns in ls_thing_value and it checks out. The only columns not being set by the constructor are `id` and `lsthing_state_id` which makes sense to me.

bffrost (approved):
LGTM

PR #209, opened 2021-05-17
747 advanced search interaction label predicates

brianbolt:
Only add label join if label predicates provided in the search criteria

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
@@ -2433,7 +2433,7 @@ public class LsThingServiceImpl implements LsThingService {
 								firstItxPredicates.add(firstThingLabelLike);
 							}
 						}else{
-							Predicate firstThingLabelLike = criteriaBuilder.like(firstThingLabel.<String>get("labelText"), '%' + interaction.getThingLabelText() + '%');
+							Predicate firstThingLabelLike = criteriaBuilder.like(criteriaBuilder.lower(firstThingLabel.<String>get("labelText")), '%' + interaction.getThingLabelText().toLowerCase() + '%');
 							firstItxPredicates.add(firstThingLabelLike);
 						}
 					}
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -2389,50 +2386,56 @@ else if (query.getRecordedDateLessThan() != null){
...
+							}else if(interaction.getOperator().equals("<=")){
+								Predicate firstThingLabelLessThan = criteriaBuilder.lessThanOrEqualTo(firstThingLabel.<String>get("labelText"), interaction.getThingLabelText());
+								firstItxPredicates.add(firstThingLabelLessThan);
+							}else{
+								Predicate firstThingLabelLike = criteriaBuilder.like(criteriaBuilder.lower(firstThingLabel.<String>get("labelText")), '%' + interaction.getThingLabelText().toLowerCase() + '%');
+								firstItxPredicates.add(firstThingLabelLike);
+							}
 						}else{
-							Predicate firstThingLabelLike = criteriaBuilder.like(criteriaBuilder.lower(firstThingLabel.<String>get("labelText")), '%' + interaction.getThingLabelText().toLowerCase() + '%');
+							Predicate firstThingLabelLike = criteriaBuilder.like(firstThingLabel.<String>get("labelText"), '%' + interaction.getThingLabelText() + '%');
```
bffrost, line 2436: Just noticed here that the analogous query above is case-insensitive but this one is case-sensitive. I think it'd make sense for this default "no operator specified" to be case-insensitive too.

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -2474,50 +2474,56 @@ else if (query.getRecordedDateLessThan() != null){
...
+								Predicate secondThingLabelGreaterThan = criteriaBuilder.greaterThanOrEqualTo(secondThingLabel.<String>get("labelText"), interaction.getThingLabelText());
+								secondItxPredicates.add(secondThingLabelGreaterThan);
+							}else if(interaction.getOperator().equals("<")){
+								Predicate secondThingLabelLessThan = criteriaBuilder.lessThan(secondThingLabel.<String>get("labelText"), interaction.getThingLabelText());
+								secondItxPredicates.add(secondThingLabelLessThan);
+							}else if(interaction.getOperator().equals("<=")){
+								Predicate secondThingLabelLessThan = criteriaBuilder.lessThanOrEqualTo(secondThingLabel.<String>get("labelText"), interaction.getThingLabelText());
+								secondItxPredicates.add(secondThingLabelLessThan);
+							}else{
+								Predicate secondThingLabelLike = criteriaBuilder.like(secondThingLabel.<String>get("labelText"), '%' + interaction.getThingLabelText() + '%');
```
bffrost, line 2520: Same thing here. Should make this case insensitive IMO.

bffrost (approved):
Left two comments that are optional to resolve. I am recommending a behavior change towards what I think is better default behavior, but feel free to opt not to change them.

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
brianbolt: Yes, I will make this change.

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
brianbolt: Interesting that its the opposite here. fixing to be the same (both lower)

PR #210, opened 2021-05-28
mcneilco/acas#756 get code name from name performance

bffrost:
Fixes mcneilco/acas#756 
- Remove unused `getCodeNameFromName(String thingType, String thingKind, String labelType, String labelKind, String json)` function.
- Total rewrite of `getCodeNameFromName(String thingType, String thingKind, String labelType, String labelKind, PreferredNameRequestDTO requestDTO)` to use CriteriaQuery and issue a single query instead of using a for loop and issuing a query for every single input.

Tested updated route with:
- thingType + thingKind + labelType + labelKind
- thingType + thingKind only
- Inputs with labelText as requestName
- Inputs with codeName as requestName
- Inputs that request the same entity, but one by label and one by codeName
- No match
- Duplicate matches

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/LsThingService.java
+++ b/src/main/java/com/labsynch/labseer/service/LsThingService.java
@@ -88,9 +88,6 @@ public interface LsThingService {
 	
 	Collection<LsThing> sortBatches(Collection<LsThing> batches);
 
-	PreferredNameResultsDTO getCodeNameFromName(String thingType,
-			String thingKind, String labelType, String labelKind, String json);
-
 	DependencyCheckDTO checkBatchDependencies(LsThing batch);
 
 	DependencyCheckDTO checkParentDependencies(LsThing parent);
--- a/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
@@ -158,6 +158,14 @@ public class LsThingServiceImpl implements LsThingService {
 		// Build the WHERE clause predicates
 		Predicate[] predicates = new Predicate[0];
 		List<Predicate> predicateList = new ArrayList<Predicate>();
+		// Thing not ignored
+		Predicate thingNotIgnored = cb.not(lsThingRoot.<Boolean>get("ignored"));
+		predicateList.add(thingNotIgnored);
+		// Label null or not ignored
+		Predicate labelIgnoredNull = cb.isNull(lsThingLabel.<Boolean>get("ignored"));
+		Predicate labelNotIgnored = cb.not(lsThingLabel.<Boolean>get("ignored"));
+		Predicate labelIgnored = cb.or(labelIgnoredNull, labelNotIgnored);
+		predicateList.add(labelIgnored);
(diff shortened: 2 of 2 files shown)


src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -183,74 +149,106 @@ public PreferredNameResultsDTO getCodeNameFromName(String thingType, String thin
...
+			// Check for multiple or zero matches
+			if (matches.size() == 1){
+				PreferredNameDTO match = matches.get(0);
+				request.setPreferredName(match.getPreferredName());
+				request.setReferenceName(match.getReferenceName());
+			} else if (matches.size() > 1){
 				responseOutput.setError(true);
 				ErrorMessageDTO error = new ErrorMessageDTO();
-				error.setLevel("error");
+				error.setLevel("MULTIPLE RESULTS");
```
brianbolt, line 241: I believe this should still be `error`

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -183,74 +149,106 @@ public PreferredNameResultsDTO getCodeNameFromName(String thingType, String thin
...
+				if (res.getRequestName().equals(request.getRequestName())){
+					matches.add(res);
+				}
+			}
+			// Check for multiple or zero matches
+			if (matches.size() == 1){
+				PreferredNameDTO match = matches.get(0);
+				request.setPreferredName(match.getPreferredName());
+				request.setReferenceName(match.getReferenceName());
+			} else if (matches.size() > 1){
```
brianbolt, line 238: Not sure it will matter, but just noticing that there is a change to the output when there are multiple matches. Previously, the service would set "" for both reference and preferred names right off the bat in this loop and now they would come back as null I believe

brianbolt (changes requested):

src/main/java/com/labsynch/labseer/service/LsThingService.java
```diff
@@ -88,9 +88,6 @@ LsThing saveLsThing(LsThing lsThing, boolean isParent, boolean isBatch,
 	
 	Collection<LsThing> sortBatches(Collection<LsThing> batches);
 
-	PreferredNameResultsDTO getCodeNameFromName(String thingType,
```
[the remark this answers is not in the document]
bffrost: No, different part of the code. I just saw duplicated code in the service and didn't find any references to this function so removed it.

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
bffrost: Yeah this was inconsistent between two sections of code. I agree it should be `error`. I'll fix.

brianbolt:
LGTM.  It's hard to think of all the scenarios when looking at all this but I think your testing is good.  This is not an issue and I think the new code handles this better, but the old code I think would not throw a Multiple results error if a code name was the same as the label text, where I believe the new code would.

PR #211, opened 2021-05-28
restoring old behavior when both a label and code name matches to not…

brianbolt:
… produce a multi match error

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
@@ -154,7 +154,6 @@ public class LsThingServiceImpl implements LsThingService {
 		CriteriaBuilder cb = em.getCriteriaBuilder();
 		CriteriaQuery<LsThing> criteria = cb.createQuery(LsThing.class);
 		Root<LsThing> lsThingRoot = criteria.from(LsThing.class);
-		Join<LsThing, LsThingLabel> lsThingLabel = lsThingRoot.join("lsLabels", JoinType.LEFT);
 
 		// Build the WHERE clause predicates
 		Predicate[] predicates = new Predicate[0];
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
```diff
@@ -177,12 +178,12 @@ public PreferredNameResultsDTO getCodeNameFromName(String thingType, String thin
 		}
 		// Label Type and Kind
 		if (labelType != null && labelType.length() > 0){
```
bffrost, line 180: Can you make the `labelNotIgnored` predicate above on lines 166-169 also use this join / `on` syntax? 
I don't think we need `labelIgnoredNull` or `labelIgnored` if we just put `labelNotIgnored` onto the join.

bffrost (changes requested):
Just one cleanup requested to put `label.ignored = False` onto the ON clause. Otherwise LGTM

src/main/java/com/labsynch/labseer/service/LsThingServiceImpl.java
brianbolt: Yes, will do

brianbolt:
Tests passed, sql produced looked like:

```
select distinct lsthing0_.id as id1_2_,
                lsthing0_.code_name as code_nam2_2_,
                lsthing0_.deleted as deleted3_2_,
                lsthing0_.ignored as ignored4_2_,
                lsthing0_.ls_kind as ls_kind5_2_,
                lsthing0_.ls_transaction as ls_trans6_2_,
                lsthing0_.ls_type as ls_type7_2_,
                lsthing0_.ls_type_and_kind as ls_type_8_2_,
                lsthing0_.modified_by as modified9_2_,
                lsthing0_.modified_date as modifie10_2_,
                lsthing0_.recorded_by as recorde11_2_,
                lsthing0_.recorded_date as recorde12_2_,
                lsthing0_.version as version13_2_
from ls_thing lsthing0_
left outer join ls_thing_label lslabels1_ on lsthing0_.id=lslabels1_.lsthing_id
and (lslabels1_.ignored<>true
     and lslabels1_.ls_type=$1
     and lslabels1_.ls_kind=$2)
where lsthing0_.ignored<>true
    and lsthing0_.ls_type=$3
    and lsthing0_.ls_kind=$4
    and (lslabels1_.label_text in ($5,
                                   $6,
                                   $7)
         or lsthing0_.code_name in ($8,
                                    $9,
                                    $10))
```

bffrost (approved):
LGTM

PR #214, opened 2021-08-11
Export SDF file from search returns ignored aliases

### LINKED ISSUE #213: Export SDF file from search returns ignored aliases
The exported aliases should not included ignored aliases

brianbolt:
Fixes #213


bffrost (approved):
LGTM

PR #221, opened 2021-11-04
Issue820

brianbolt:
I tested this on a server and verified it works correctly by running this curl which was previously throwing a null pointer exception.

```
curl http://localhost:3001/api/getContainerLocationTree
```


bffrost (approved):
LGTM, much cleaner

PR #224, opened 2021-11-18
Acas 153

brianbolt:


src/main/resources/db/migration/postgres/R__check_for_standardizer_configuration_changes.java
```diff
@@ -0,0 +1,346 @@
...
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.RowMapper;
+
+import com.labsynch.labseer.domain.StandardizationHistory;
+import com.labsynch.labseer.dto.configuration.MainConfigDTO;
+import com.labsynch.labseer.dto.configuration.StandardizerSettingsConfigDTO;
+import com.labsynch.labseer.utils.Configuration;
+
+public class R__check_for_standardizer_configuration_changes implements SpringJdbcMigration, MigrationChecksumProvider {
```
bffrost, line 22: @brianbolt I thought you had moved away from the repeatable flyway migration. Is it on purpose that this is still in here?

brianbolt: Thanks, just removed it.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+					qcCompound.setChangedStructure(true);
+					for (int hit:hits){
+						List<StandardizationDryRunCompound> searchResults = StandardizationDryRunCompound.findStandardizationDryRunCompoundsByCdId(hit).getResultList();
+						for (StandardizationDryRunCompound searchResult : searchResults){
+							if (searchResult.getCorpName().equalsIgnoreCase(qcCompound.getCorpName())){
+								qcCompound.setChangedStructure(false);
+								newDupeCount = newDupeCount-1;
+							} else {
+								if (StringUtils.equals(searchResult.getStereoCategory(), qcCompound.getStereoCategory())
+										&& StringUtils.equalsIgnoreCase(searchResult.getStereoComment(), qcCompound.getStereoComment())){
```
bffrost, line 268: Does this logic handle cases when one or both are `null`? I recall in other code (probably in BulkLoadServiceImpl) we implemented some utilities or shortcuts for this "is the parent a dupe" logic.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+						}
+					}
+					
+					hits = chemStructureService.searchMolStructures(qcCompound.getMolStructure(), StructureType.PARENT, SearchType.DUPLICATE_TAUTOMER);
+					oldDuplicateCount = hits.length;
+					for (int hit:hits){
+						List<Parent> searchResults = Parent.findParentsByCdId(hit).getResultList();
+						for (Parent searchResult : searchResults){
+							if (searchResult.getCorpName().equalsIgnoreCase(qcCompound.getCorpName())){
+								oldDuplicateCount = oldDuplicateCount-1;
```
bffrost, line 288: I think the "changedStructure" logic should be here instead.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+		}
+		standardizationHistory.setStandardizationStart(new Date());
+		standardizationHistory.setStandardizationComplete(null);
+		standardizationHistory.setStandardizationStatus("running");
+		standardizationHistory.persist();
+		
+		Integer result;
+		try {
+			result = this.runStandardization();
+		} catch (CmpdRegMolFormatException | IOException | StandardizerException e) {
```
bffrost, line 420: Does this mean a single bad structure will cause the whole restandardization to fail?
Just want to understand the expected behavior here.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+		List<StandardizationSettings> standardizationSettingses = StandardizationSettings.findAllStandardizationSettingses("modifiedDate", "DESC");
+		StandardizationSettings standardizationSettings = new StandardizationSettings();
+		if(standardizationSettingses.size() > 0) {
+			standardizationSettings = standardizationSettingses.get(0);
+		}
+		return(standardizationSettings);
+	}
+
+	@Override
+	public List<StandardizationHistory> getStanardizationHistory() {
```
bffrost, line 471: nit: typo in function name

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+	
+	@Transactional
+	private int runDryRun() throws CmpdRegMolFormatException, IOException, StandardizerException {
+		logger.info("standardization dry run initialized");
+		logger.info("step 1/3: resetting dry run table");
+		this.reset();
+		logger.info("standardization dry run 1% complete");
+		logger.info("step 2/3: populating dry run table");
+		int numberOfDisplayChanges = this.populateStanardizationDryRunTable();  
+		logger.info("standardization dry run 20% complete");
```
bffrost, line 506: 20% figure here is a bit strange. but I see the logging during populating the dry run table is separate. Did this come from an estimate that the restandardization takes 20% of the time and dupe checking takes 80% of the time?

There is already logging of % progress within step 2 and step 3. I think the "step 1/3, step 2/3" lines make sense, but the percentages don't make sense to me in this function. 1%, 20%, done?

bffrost (changes requested):
Did my best on this large PR. Nits are optional, but please respond to the other comments.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -0,0 +1,128 @@
...
+	private String molStructure;
+
+	private String comment;
+
+	private boolean ignore;
+
+	public StandardizationDryRunCompound() {
+	}
+
+	public static List<Long> getAllIds() {
```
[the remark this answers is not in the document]
brianbolt: Yes, this was unused, removed.

src/main/java/com/labsynch/labseer/dto/configuration/ServerSettingsConfigDTO.java
```diff
@@ -0,0 +1,70 @@
+package com.labsynch.labseer.dto.configuration;
+
+import org.springframework.roo.addon.javabean.RooJavaBean;
+import org.springframework.roo.addon.json.RooJson;
+import org.springframework.roo.addon.tostring.RooToString;
+
+@RooJavaBean
+@RooToString
+@RooJson
+public class ServerSettingsConfigDTO {
```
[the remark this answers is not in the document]
brianbolt: Yes, didn't noticed it was re-introduced. Re-removed.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+			}
+		} catch (StandardizerException e) {
+			logger.error("Caught error on checking standardization state", e);
+		}
+	}
+
+
+	@Transactional
+	@Override
+	public int populateStanardizationDryRunTable() throws CmpdRegMolFormatException, IOException, StandardizerException{
```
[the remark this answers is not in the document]
brianbolt: Fixed file by reformatting but it used the this indentation, not the others so these remained the same.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Yes, if any one of the exceptions are thrown here, then it will stop and set the standardization to fail.  This of course depends on the underlying code not swallowing any exceptions so I 'll need to make sure they throw when encounting these exceptions.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+				boolean asDrawnDisplaySame = chemStructureService.isIdenticalDisplay(asDrawnStruct, stndznCompound.getMolStructure());
+				if (!asDrawnDisplaySame){
+					stndznCompound.setAsDrawnDisplayChange(true);
+					logger.debug("the compounds are NOT matching: " + parent.getCorpName());
+					nonMatchingCmpds++;
+				}
+			}
+			cdId = chemStructureService.saveStructure(stndznCompound.getMolStructure(), StructureType.DRY_RUN, false);
+			if (cdId == -1){
+				logger.error("Bad molformat. Please fix the molfile: " + stndznCompound.getMolStructure());
```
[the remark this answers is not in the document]
brianbolt: I added this.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+					logger.debug("mol has a weight of 0 - skipping");			
+				} else {
+					hits = chemStructureService.searchMolStructures(qcCompound.getMolStructure(), StructureType.DRY_RUN, SearchType.DUPLICATE_TAUTOMER);
+					newDupeCount = hits.length;
+					qcCompound.setChangedStructure(true);
+					for (int hit:hits){
+						List<StandardizationDryRunCompound> searchResults = StandardizationDryRunCompound.findStandardizationDryRunCompoundsByCdId(hit).getResultList();
+						for (StandardizationDryRunCompound searchResult : searchResults){
+							if (searchResult.getCorpName().equalsIgnoreCase(qcCompound.getCorpName())){
+								qcCompound.setChangedStructure(false);
```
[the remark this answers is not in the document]
brianbolt: I updated qcCompound to be dryRun compound and moved the changedStructure logic to the section below.  The logic now is that a changed structure is one that cannot be found any long in the parent structure table when doing a duplicate tautomer search.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Yes, thanks. Moved it here instead.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: I removed these lines.  They were from some timings I did long ago when using jchem.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Yes, null is handled by the StringUtils functions

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: I just caused a lot of diffs but to fix this i ran a format using my vs code formatter.  There were lots of indentation problems with this file.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+			int[] hits;	
+			StandardizationDryRunCompound qcCompound;
+			String newDuplicateCorpNames = "";
+			String oldDuplicateCorpNames = "";
+			int newDupeCount = 0;
+			int oldDuplicateCount = 0;
+			
+			float percent = 0;
+			int p = 1;
+		    float previousPercent = percent;
```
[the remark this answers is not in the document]
brianbolt: I just caused a lot of diffs but to fix this i ran a format using my vs code formatter. There were lots of indentation problems with this file.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -0,0 +1,524 @@
...
+		List<Lot> queryLots;
+		Integer runNumber = StandardizationDryRunCompound.findMaxRunNumber().getSingleResult();
+		if (runNumber == null){
+			runNumber = 1;
+		} else {
+			runNumber++;
+		}
+		float percent = 0;
+		int p = 1;
+	    float previousPercent = percent;
```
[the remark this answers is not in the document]
brianbolt: I just caused a lot of diffs but to fix this i ran a format using my vs code formatter. There were lots of indentation problems with this file.

brianbolt:
@bffrost Besides the issues you noted. I found had to refactor RDkit structure a bit to have it extend and AbstractRDkit structure.   The reason for this is because if I used a non abstract class, then hibernate will find all subclasses of rdkitstructure.

i.e. during restandardiation when I had the same structure in the parent structure table (rdkit structure) and the dry run structure table (dry run structure).  The following line would search and find both rows in the seperate tables:

https://github.com/mcneilco/acas-roo-server/blob/83b08e62c723498e6ecea1473cd5879365a24de8/src/main/java/com/labsynch/labseer/chemclasses/rdkit/ChemStructureServiceRDKitImpl.java#L393

This is a pretty cool feature but terrible for our usage.  By having an abstract class and extending it for each subclass, this fixes the problem I was seeing...took me a while to find this one.

src/main/java/com/labsynch/labseer/domain/RDKitStructure.java
```diff
@@ -18,42 +8,7 @@
...
-    @DateTimeFormat(style="M-")
-    private Date recordedDate;
-
-    public void updateStructureInfo(String mol, String reg, String preReg) {
-        this.setMol(mol);
-        this.setReg(reg);
-        this.setPreReg(preReg);
-    } 
+@RooJpaActiveRecord(sequenceName = "RDKIT_STRUCTURE_PKSEQ", finders = { "findRDKitStructuresByRegEquals", "findRDKitStructuresByPreRegEquals"})
+public class RDKitStructure extends AbstractRDKitStructure {
```
bffrost, line 12: Now that you have `AbstractRDKitStructure` and the specific classes seem to extend that, what is the purpose of this `RDKitStructure` class?

bffrost: Ah, I see this is the class name for the parent structure. I think it would be clearer to rename this class `RDKitParentStructure` to make it crystal clear.

src/main/java/com/labsynch/labseer/chemclasses/rdkit/ChemStructureServiceRDKitImpl.java
```diff
@@ -268,11 +269,24 @@ public RDKitStructure getRDKitStructureFromService(String molfile) throws CmpdRe
 
 	@Override
 	public int saveStructure(String molfile, StructureType structureType, boolean checkForDupes) {
+
+		Long cdId=0L;
 		try {
 			RDKitStructure rdkitStructure = getRDKitStructureFromService(molfile);
```
bffrost, line 275: Should this be an `AbstractRDKitStructure` ? It seems like it.

src/main/java/com/labsynch/labseer/chemclasses/rdkit/ChemStructureServiceRDKitImpl.java
```diff
@@ -268,11 +269,24 @@ public RDKitStructure getRDKitStructureFromService(String molfile) throws CmpdRe
...
 			if (structureType == StructureType.PARENT){
-				rdkitStructure.persist();
+				// Can't type cast from subclass to superclass so we go to json and back
+				RDKitStructure rdStructure = RDKitStructure.fromJsonToRDKitStructure(rdkitStructure.toJson());
+
+				if(checkForDupes){
+					List<RDKitStructure> rdkitStructures =  RDKitStructure.findRDKitStructuresByRegEquals(rdStructure.getReg()).getResultList();
+					if(rdkitStructures.size() > 0){
+						logger.error("RDkit parent structure already exists with id "+ rdkitStructures.get(0).getId());
+						return 0;
```
bffrost, line 285: This is an unexpected case, right? Since we expect the registration code above will catch this and not ask to `saveStructure`?

I think it would be safer to throw an exception rather than return 0, to make this break visibly in this unexpected state.

bffrost (changes requested):
New changes are looking good. Noted a few more points, but once those are resolved this will be good to merge.

src/main/java/com/labsynch/labseer/chemclasses/rdkit/ChemStructureServiceRDKitImpl.java
brianbolt: no, you can't new up an abstract class so you have to pick one of the classes that implements it.

src/main/java/com/labsynch/labseer/chemclasses/rdkit/ChemStructureServiceRDKitImpl.java
brianbolt: I actually did try to find the right thing to do here and saw this:
https://github.com/mcneilco/acas-roo-server/blob/c942475947c904a7aa470828eb370449d0418f3b/src/main/java/com/labsynch/labseer/service/SaltFormServiceImpl.java#L160

And jchem does the same:
https://github.com/mcneilco/acas-roo-server/blob/7ecf6d6581af7286b55c0b5455ecff334db46f33/src/main/java/com/labsynch/labseer/chemclasses/jchem/ChemStructureServiceJChemImpl.java#L349

I am going to update the logger line above to be the following because its not always parent structure in this function:
```
						logger.error("RDkit structure for " + structureType + " type already exists with id "+ rdkitStructures.get(0).getId());
```

src/main/java/com/labsynch/labseer/domain/RDKitStructure.java
brianbolt: That makes sense, however, if I make that change now, its going to make for some annoying conflicts as part of my next pull request.  I'd prefer to make this change when I rename everything to be bbchem rather than rdkit.

PR #228, opened 2021-12-04
ACAS-164

brianbolt:
Its not possible to inject the configurations into flyway currently and its probably better to save the salt structures as part of standardization anyway instead of via flyway.  I added a function to check for and save any missing mol structures.


bffrost (approved):
LGTM, this seems like a reasonable place to fix this.

PR #234, opened 2021-12-16
Switch to repo.spring.io

brianbolt:


pom.xml
```diff
@@ -57,36 +57,38 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.]]>
 	</scm>
 	<repositories>
 		<repository>
-			<id>spring-maven-release</id>
-			<name>Spring Maven Release Repository</name>
-			<url>https://maven.springframework.org/release</url>
+			<id>spring-snapshots</id>
+			<name>Spring Snapshots</name>
+			<url>https://repo.spring.io/snapshot</url>
```
bffrost, line 62: @brianbolt Do we want snapshot rather than release? Not sure why this was switched

brianbolt: Yeah, we do want release. I copy pasted this and missed that it changed.

brianbolt: I've reverted this for now and will come back to it later since the old builds are working

bffrost: Sounds good, thanks

PR #236, opened 2021-12-21
Acas 3 misc

brianbolt:


brianbolt:
Main changes here are:

1. Make re-standardization non transactional to flush sessions to boost performance
2. Revert jchem impl changes made during creation of bbchem impl
3. Actually re-standardize structures

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -145,10 +153,21 @@ public void onApplicationEvent(ContextRefreshedEvent event) {
 	}
 
 	@Transactional
+	private int saveDryRunStructure(String molStructure) {
+		int cdId = chemStructureService.saveStructure(molStructure, StructureType.DRY_RUN, false);
+		return cdId;
+	}
+
 	@Override
 	public int populateStanardizationDryRunTable()
```
bffrost, line 162: nit: still a typo here in `Stanardization`. Missing a `d`

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -212,30 +231,40 @@ public int populateStanardizationDryRunTable()
...
 			previousPercent = percent;
 			p++;
+
 		}
 		logger.info(
 				"total number of non matching; structure, display or as drawn display changes: " + nonMatchingCmpds);
 		return (nonMatchingCmpds);
 	}
 
-	@Transactional
```
bffrost, line 238: Curious about why @Transactional was removed on this function, but maybe was not necessary if the aliases are already populated on the `parent` object.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -397,43 +440,82 @@ public void reset() {
 		StandardizationDryRunCompound.truncateTable();
 	}
 
+	@Transactional
+	public Boolean updateStructure(String molStructure, int cdId) {
```
bffrost, line 444: nit: Is there a reason why you didn't just make `chemStructureService.updateStructure` `@Transactional`?
This wrapper does almost nothing, seems like it'd make more sense to just make the inner function itself transactional.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -397,43 +440,82 @@ public void reset() {
...
+				logger.warn("Could not update structure for parent: " + parentId + "  " + parent.getCorpName());
+				logger.info("Assuming the structure did not exist in the first place and saving a new one");
+				int newCdId = chemStructureService.saveStructure(standardizedMol, StructureType.PARENT, false);
+				parent.setCdId(newCdId);
+				logger.info("Updated parent with new cdId: " + newCdId);
+			}
+
+			// Save the standardized structure and possibly the new cdid to the parent
+			parent.setMolStructure(standardizedMol);
+			parent.persist();
```
bffrost, line 499: This maybe should be `parent.merge()` since the parent is not new

bffrost (changes requested):
Mostly looks good. Just a few minor things to address. I think the `persist` vs `merge` one does need to be changed.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: I wasn't sure how making that specific service transactional for everyone would change things so I chose the low risk change to make JUST the standardization service portion transactional.

PR #239, opened 2022-01-03
Feature/batches codetable

brianbolt:


brianbolt:
@philippcheung said LGTM offline

PR #240, opened 2022-01-06
Brianbolt/issue232

brianbolt:
Apparently all pom files inherit from the maven super [POM](https://maven.apache.org/ref/3.0.4/maven-model-builder/super-pom.html) and because the main pom.xml file has repo.maven.apache.org in it as `<id>central</id>` our builds will continue to pull from this repo.  I tried removing this by setting our repo ids to "central" but it failed to find some depdencies.  So ultimately we want to also pull from repo.maven.apache.org.


bffrost (approved):
LGTM

PR #241, opened 2022-01-12
Brianbolt/issue848

brianbolt:


bffrost (approved):
LGTM. Implementation makes sense, it's easy to read, and I support your decision not to implement this function in the other engines right now.

PR #242, opened 2022-01-13
Brianbolt/issue849

brianbolt:


bffrost (approved):
LGTM, but I had one question / doubt. The SQL to update lot MW depends on the updates to parent MW having been committed to the DB, and I don't see an explicit `flush` or the end of a transactional block between when we do `parent.merge()` and when we run `restandardizeLots`.

It's entirely possible this is working as intended, but I'd rest easier if you can confirm you've run some "double-check" SQL on a test server after this and confirmed that the lot MWs are all updated as expected.

brianbolt:
Yes, I had this same question/doubt and did the double check to confirm the weights were updated properly.

PR #248, opened 2022-01-15
Brianbolt/issue247

brianbolt:
This is just the services around the new query interface.


src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -114,6 +124,99 @@ public StandardizationHistory fetchStats() {
...
+
+		// As drawn display change
+		if (dryRunSearch.getAsDrawnDisplayChange() != null) {
+			predicates.add(criteriaBuilder.equal(root.get("asDrawnDisplayChange"), dryRunSearch.getAsDrawnDisplayChange()));
+		}
+
+		if(predicates.size() > 0) {
+			Predicate[] predicatesToAdd = new Predicate[0];
+			predicatesToAdd = predicates.toArray(predicatesToAdd);
+			Predicate firstItxAndPredicates = criteriaBuilder.and(predicatesToAdd);
```
bffrost, line 210: nit: copypasta "firstItx" in the variable name doesn't belong.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -114,6 +124,99 @@ public StandardizationHistory fetchStats() {
...
+			criteria.where(predicates.toArray(new Predicate[predicates.size()]));
```
bffrost, line 211: Something looks off here. You're not using that `firstItxAndPredicates` variable you just created. Shouldn't this line just be `criteria.where(firstItxAndPredicates)` or something?

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -114,6 +124,99 @@ public StandardizationHistory fetchStats() {
...
+		if (dryRunSearch.getDeltaMolWeight() != null) {
+			predicates = buildPredicateFromNumericValue(criteriaBuilder, root, predicates, "deltaMolWeight", dryRunSearch.getDeltaMolWeight().getValue(), dryRunSearch.getDeltaMolWeight().getOperator());
+		}
+
+		// Old mol weight
+		if (dryRunSearch.getOldMolWeight() != null) {
+			predicates = buildPredicateFromNumericValue(criteriaBuilder, root, predicates, "oldMolWeight", dryRunSearch.getOldMolWeight().getValue(), dryRunSearch.getOldMolWeight().getOperator());
+		}
+
+		// Corp name in list
```
bffrost, line 167: The one thing I see missing based on the requirements is the ability to _exclude_ corp names. I was picturing a list of corp names, then a boolean that controls whether the list gets _included_ or _excluded_

bffrost (changes requested):
Found one missing bit and one weird looking line of code. Otherwise it makes sense.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
brianbolt: Thanks, just added that Boolean option

bffrost (approved):
LGTM

PR #261, opened 2022-01-27
Update to new tomcat image which has openssl and cacerts permissions for runner

### LINKED ISSUE #260: Add openssl and allow runner user to import a cert to cacerts
In order to authenticate against ldap's using ssl and self signed certificates, a command is added to the beginning of the normal "catalina.sh run" command.  The command uses openssl to download the cert and then keytool to import the cert into the jvm keystore.  In order to do this, openssl must be installed and the runner user must have the ability to import the keytool.  In the 1.13.7 update we updated to the centos image which does not have openssl by default and we switched to a non root "runner" user.  This ticket is to add openssl to the 1.13.7 images and give the runner user the ability to import the keystore.

brianbolt:
I tested running the OpenSSL and keytool import command on a server which used self signed certs.

fixes #260

See pull request https://github.com/mcneilco/tomcat-maven/pull/1


bffrost (approved):
LGTM

PR #264, opened 2022-01-28
Fixing jchem and indigo builds

brianbolt:
These are additionally changes I missed for the other chemistry packages.  These are required methods on CmpdRegMolecule.


bffrost (approved):
LGTM

Issue #273, opened 2022-02-04
BBChem Creg: Don't check for dupe structures when standardization or registration error


brianbolt:
If a structure fails to standardize or a structure fails to register, we should not search for duplicates because either the structure is not truly standardized or all null registration compounds will be returned by the search method resulting in long detected duplicate lists if there are lots of error compounds.

PR #274, opened 2022-02-04
Brianbolt/issue272

brianbolt:
Tested on a system and verified that the bbchem_standardization_dry_run_structure table was populated during standardization rather than the bbchem_dry_run_structure


bffrost (approved):
Approving without a close look since this has the same changes as #273  which I carefully reviewed.

PR #275, opened 2022-02-04
Brianbolt/issue273

### LINKED ISSUE #273: BBChem Creg: Don't check for dupe structures when standardization or registration error

brianbolt:
Fixes #273


bffrost (approved):
LGTM! Thanks for the fast turnaround here.

PR #294, opened 2022-03-16
Docker build roo throws unable to convert uid/gid chown string to host mapping

### LINKED ISSUE #293: Docker build roo throws unable to convert uid/gid chown string to host mapping
Docker build of roo throws the following error:
 
```
unable to convert uid/gid chown string to host mapping: can't find uid for user runner: no such user: runner
```
The error occurs on this line of the docker file:
```
ADD --chown=runner:runner pom.xml /src/pom.xml 
```
This is because the runner user doesn't exist at this point during the build.  This doesn't occur when using docker build kit but we aren't running that on the server.
 
I tested removing --chown=runner:runner at that step and it fixes the issue.  However, I get another error:
 
```
When using COPY with more than one source file, the destination must be a directory and end with a /
```

Caused by this line:
```
COPY --chown=runner:runner --from=compile /src/target/acas* /usr/local/tomcat/webapps/acas
```

brianbolt:
## Description
Removed chown statement as it was unnecessary and caused issues when using docker build without docker build kit
Added "/" to the end of copy step of acas to fix and issue when building without docker build kit

## Related Issue
Fixes #293

## How Has This Been Tested?
I tested a non docker build kit build of the Jchem version and verified the server came up and ran
I tested a docker build kit build of the bbchem version and verified the server came up and ran

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ADD 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
 RUN     mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
 
 FROM 	${CHEMISTRY_PACKAGE} as compile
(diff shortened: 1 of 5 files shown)


bffrost (changes requested):
@brianbolt you've checked in the jchem jar accidentally again. Please remove, otherwise looks good.

PR #322, opened 2022-05-06
Allow SEL upload w/ parent alias values within lot corp name

### LINKED ISSUE #321: Allow SEL upload w/ parent alias values within lot corp name
Please implement an enhancement to ACAS that allows lot corp names to be formed from a parent alias and lot number, in additional support to the standard corp name and lot number.

brianbolt:
## Description
Read new acas config `server.service.external.preferred.batchid.allowParentAliasLotNames` and lookup Lot Corp Names by `Parent Alias Lot Corp Name`

See mcneilco/acas#942 for feature description

## Related Issue
Fixes #321 

## How Has This Been Tested?
See mcneilco/acas#942 for testing details


src/main/java/com/labsynch/labseer/api/ApiPreferredIdController.java
```diff
@@ -32,7 +37,7 @@ public ResponseEntity<String> jsonArrayCheckCorpNameExists(@RequestBody String j
 		HttpHeaders headers = new HttpHeaders();
 		headers.add("Content-Type", "application/json");
 		Collection<PreferredNameDTO> preferredNameDTOs = PreferredNameDTO.fromJsonArrayToPreferredNameDTO(json);
-		preferredNameDTOs = PreferredNameDTO.getPreferredNames(preferredNameDTOs);
```
brianbolt, line 35: I don't think this every really belonged on the DTO.  However, the real reason I moved it to lotService is because I needed access to the @Autowired properties utils service to be able to get access to the various acas configs during the requested name parsing.

src/main/java/com/labsynch/labseer/service/LotCorpName.java
```diff
@@ -0,0 +1,42 @@
+package com.labsynch.labseer.service;
+
+public class LotCorpName {
```
brianbolt, line 3: I created this utility class as I needed a 3 element item and didn't want to hack it with something weird like an array or `HashMap<String, HashMap<integer,String>>`. Might be a little overkill but its a nice utility for this stuff.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -317,6 +335,107 @@ public String generateParentLotName(Lot lot) {
...
+		}
+		if (longestSub.length() > 0) {
+			return Integer.parseInt(longestSub);
+		} else {
+			return -1;
+		}
+	}
+
+	@Override
+	public LotCorpName splitCorpNameComponents(String inputString) {
```
brianbolt, line 360: This is the most complicated piece all due to not wanting to abandon the rarely used `appendSaltCodeToLotName` feature.  This code is essentially the reverse of `generateLotNameFromLotCorpName`

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -285,12 +292,23 @@ public String appendSaltCode(String corpName, Lot lot) {
 
 	@Override
 	public String generateParentLotName(Lot lot) {
```
brianbolt, line 294: I was going to approach this problem differently at first and refactored the `generateParentLotName` method to fit that approach.  I changed course but left this refactor in because I thought it could be beneficial and is clearer.

brianbolt (commented):
The main part of the logic of this new code:

 - Split the requested name into the expected format components
   - corp name (but parent alias in this case)
   - batch
   - salt (rare case but supported by this)
 - Return "" if the string could not be parsed to match the correct format
 -  Search for parent aliases that match
    - return "" if multiple or zero parents match
    - return lot corp name if lot number matches the single found parent

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
bffrost: I'm concerned about the reliance on separators, plus the fact that our configs allow empty-string as a separator.
I'm having trouble thinking of efficient alternative approaches. The simplest approach would be for us to have a cache of alternate LotCorpNames that utilize the aliases, then search against that.

Two big picture alternatives to your current approach:
- Add the necessary java code to create and save Lot Aliases based on incoming Parent Aliases whenever new Lots or Parent Aliases are registered. Also add a flyway migration to backfill these. Then simply search the input string directly against Lot Alias.
- Without using Lot Alias, an alternative searching approach:
 -- Fetch all Parent Aliases in the DB, scan them against the incoming name for substring matches
 -- For any matching parents, get the full LotCorpNames that exist in the DB, then do a string swap (find/replace) of parent corp name -> parent alias within the LotCorpName, and check if it fully matches the incoming string.

bffrost (changes requested):

src/main/java/com/labsynch/labseer/service/LotCorpName.java
```diff
@@ -0,0 +1,42 @@
...
+    public String getSaltCode() {
+        return this.saltCode;
+    }
+
+    public void setSaltCode(String saltCode) {
+        this.saltCode = saltCode;
+    }
+
+    @Override
+    public String toString() {
```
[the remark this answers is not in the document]
brianbolt: Its just a utility class which isn't expected to create a concatenated I'm removing this since it's not being used.

src/main/java/com/labsynch/labseer/api/ApiPreferredIdController.java
brianbolt: Yeah make sense. Thinking about it a bit more, I think I like option 2 more.  Mainly because it's less maintenance of data.  Manage all the updates deletes and additions to lot aliases when parent aliases change seems like a bit of a headache.

brianbolt: Worked on this a bit but I realize now we have an issue when the lot corp name e.g. "LIT-00000001-0001" has no connection to the parent corp name which is the case on some system.  I think the best approach might be your second approach to finding the parent alias, combined with my initial abandoned approach (which lead to the refactor of the `generateParentLotName` in this PR.

-- Fetch all Parent Aliases in the DB, scan them against the incoming name for substring matches
-- From the matching parents, create a "Parent Alias Lot Corp Name" by passing in the alias as the "parent corp name" and the lot. which will generate what should be a matching string to the input string the user passed in.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
bffrost: Makes sense to me. Please proceed with the approach you've outlined here.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -285,12 +292,23 @@ public String appendSaltCode(String corpName, Lot lot) {
...
 		logger.debug("generating the new lot corp name");
 		String corpName = lot.getSaltForm().getParent().getCorpName();
 		logger.debug("Parent corpName = " + corpName);
 		int lotNumber = this.generateParentLotNumber(lot);
 		lot.setLotNumber(lotNumber);
 
+		return (generateLotNameFromLot(lot));
+	}
+
+	public String generateLotNameFromLot(Lot lot) {
```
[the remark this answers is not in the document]
brianbolt: Thanks, refactored this with your suggestions.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
brianbolt: This worked out, I just committed the changes and submitting for re-review now.

bffrost (approved):
LGTM. Only question remaining is whether we should leave in the unused`splitCorpNameComponents` utility or just remove it.
I'm inclined to remove it since it would be tempting to use and we'd just run into those issues with empty-string separators. So removing it will make it harder for us to trip over that in the future.

brianbolt:
Thanks, I mean to remove it but forgot. Just removed it and the remaining unused methods it depended on.

PR #326, opened 2022-05-11
Update default stereo categories on new systems

brianbolt:
## Description
This is a flyway migration to change the stereo categories on a system which has no parents (i.e. a new system).

The new categories are:
- No stereochemistry
- Single stereoisomer
- Single stereoisomer - arbitrary assign
- Single stereoisomer - partial assign
- Mixture
- Mixture - arbitrary assign
- Mixture - partial assign
- Relative mixture
- Relative mixture - arbitrary assign
- Unknown

## Related Issue
ACAS-89

## How Has This Been Tested?
- Updated an existing system and confirmed stereo categories were unchanged by checking the table and verified this log line was printed:

```
roo_1        | 11 May 2022 20:04:37 3013 [main] INFO  com.labsynch.labseer.db.migration.postgres.V2_3_0_8__update_stereo_category_defaults [] - Not updating stereo category defaults because there are already rows in the parent table
```

- Created a fresh system (empty database) and confirmed that the new stereo categories were present.
- Verified that the single reg drop down showed the new stereo categories and that I could register a compound with a new stereo category.

### DIFF
--- a/src/main/java/com/labsynch/labseer/db/migration/postgres/V2_3_0_8__update_stereo_category_defaults.java
+++ b/src/main/java/com/labsynch/labseer/db/migration/postgres/V2_3_0_8__update_stereo_category_defaults.java
@@ -1,0 +1,48 @@
+package com.labsynch.labseer.db.migration.postgres;
+
+import org.flywaydb.core.api.migration.spring.SpringJdbcMigration;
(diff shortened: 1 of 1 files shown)


bffrost (approved):
LGTM

PR #327, opened 2022-05-18
ACAS-98: Make SDF bulk loader case insensitive

bffrost:
## Description
- Add ignoreCase lookups for vendor, physical state, stereo category
- I wasn't totally sure how far to take this. There are some lesser-used categorical lookups that this could be extended to, like Parent Annotation

## How Has This Been Tested?
Ran acasclient tests including new one for wrong-case inputs: https://github.com/mcneilco/acasclient/pull/55

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiPhysicalStateController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiPhysicalStateController.java
@@ -1,8 +1,15 @@
 package com.labsynch.labseer.api;
 
+import java.util.ArrayList;
+import java.util.List;
+
+
 import com.labsynch.labseer.domain.PhysicalState;
+import com.labsynch.labseer.service.ErrorMessage;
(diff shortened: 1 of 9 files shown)


brianbolt:
Wondering if we should do something like this:

https://stackoverflow.com/questions/4124185/postgresql-unique-indexes-and-string-case

bffrost:
> 

@brianbolt to my eye, those lines are just checking for when a `null` or empty string `code` is passed in. They don't do case-insensitive dupe checks.
I like the suggestion of case-insensitive unique checks. I think to fully switch over we'll need to fix a few more places:
- Also apply the change to DDict Values since some properties such as additional scientist are already implemented as DDicts which have the same issue.
- Add flyway migrations that clean up duplicate values before applying the new unique indexes
- Update the uniqueness check in the CodeTableAdmin UI modules to also be case-insensitive

I'll start working on those.

brianbolt (approved):
LGTM

PR #333, opened 2022-06-03
Return bulk load file id

brianbolt:
## Description
Add bulk load file `id` to the output of bulk load reg responses

## Related Issue
ACAS-311
ACAS-273

## How Has This Been Tested?
Wrote ACAS client tests for ACAS-273 and changed the behavior of the purge bulk load test to use the `id` as a handle when bulk loading and then purging a file.


bffrost (approved):
LGTM

PR #335, opened 2022-06-14
ACAS-316 add maven build caching

bffrost:
## Description
Add maven build caching to multistage builds file. Item #3 from https://www.baeldung.com/ops/docker-cache-maven-dependencies

## How Has This Been Tested?
Ran builds against indigo, switched branches, reran build and it went super fast.
Switched to building bbchem, reran build and it went super fast too.

Merging this to the release will test the CI builds.

### DIFF
--- a/Dockerfile-multistage
+++ b/Dockerfile-multistage
@@ -17,9 +17,9 @@ FROM 	${CHEMISTRY_PACKAGE} as compile
 
 ADD 	pom.xml /src/pom.xml
 WORKDIR /src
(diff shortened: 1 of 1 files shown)


bffrost:
I think this caching will not help CI builds but I think it won't hurt them either since the cache will get wiped out when the build is complete.
There is a way in GitHub Actions to store cache artifacts between builds but it would substantially increase the storage used by the free tier of GitHub so I don't think it's worth it just to speed up roo builds in CI.

brianbolt (approved):
LGTM

PR #336, opened 2022-06-27
Add charset=utf-8 to all json responses

brianbolt:


bffrost (approved):
LGTM. Seeing this change made everywhere does bring to mind the idea of refactoring these to use a common utility function like `getJsonHeaders` or something, but not required for this PR.

PR #338, opened 2022-06-28
ACAS-327: Add catch for null property value

bffrost:
## Description
We hit a NullPointerException in the BBChem ACAS code when it encounters an empty SDF property. This should fix it.

## How Has This Been Tested?
Not tested


brianbolt:
LGTM.  Assuming some testing will be done before merging but. the change is safe anyway.

PR #339, opened 2022-06-28
Acas 320 bb

brianbolt:
## Description

 - Refactored registerSDF main loop into `@Transactional(propagation = Propagation.REQUIRES_NEW)` so that any  single compound is atomic but the main loop continues on sql error.
  - For standardization implementation `@Transactional(propagation = Propagation.REQUIRES_NEW)` to reduce batch size of main transactions.
   - https://docs.spring.io/spring-framework/docs/current/reference/html/data-access.html#tx-propagation
 
## Related Issue
ACAS-320

## How Has This Been Tested?
Added acasclient test for large (1K) compound load including a compound which errors
Due to Indigo chemistry (open source) not currently implementing standardization I couldn't add acasclient tests for standardization. Instead I used a bbchem server to run dry run standardization. Additionally, I truncated the bbchem_parent_structure table and ran dry/run and full parent restandardization to verify speeds.
Verified that duplicate compounds within the same file are detected via `test_047_load_sdf_with_salts` acasclient tests
Ran all acas client tests

### DIFF
--- a/src/main/java/com/labsynch/labseer/chemclasses/indigo/ChemStructureServiceIndigoImpl.java
+++ b/src/main/java/com/labsynch/labseer/chemclasses/indigo/ChemStructureServiceIndigoImpl.java
@@ -12,6 +12,7 @@ import java.util.Map;
 import java.util.Set;
 
 import javax.persistence.EntityManager;
+import javax.persistence.PersistenceException;
 import javax.persistence.Query;
 
 import org.apache.commons.lang3.ArrayUtils;
@@ -276,7 +277,7 @@ public class ChemStructureServiceIndigoImpl implements ChemStructureService {
 
 			return hitList;
 
-		} catch (JpaSystemException sqlException) {
+		} catch (JpaSystemException | PersistenceException sqlException) {
 			logger.error("Caught search error", sqlException);
 			Exception rootCause = new Exception(ExceptionUtils.getRootCause(sqlException).getMessage(),
 					ExceptionUtils.getRootCause(sqlException));
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -522,11 +522,7 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 					numNewLotsOldParentsLoaded++;
 				writeRegisteredMol(numRecordsRead, mol, metalotReturn, mappings, registeredMolExporter,
 						registeredCSVOutStream, isNewParent, results);
-				if (numRecordsRead % 100 == 0) {
-					logger.info("flushing bulk loader session");
-					session.flush();
-					session.clear();
-				}
+
 				currentTime = new Date().getTime();
 				if (currentTime > startTime) {
 					logger.info("SPEED REPORT:");
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -20,6 +20,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.persistence.NoResultException;
+import javax.persistence.PersistenceException;
 
 import com.labsynch.labseer.chemclasses.CmpdRegMolecule;
 import com.labsynch.labseer.chemclasses.CmpdRegMoleculeFactory;
@@ -80,6 +81,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.TransactionSystemException;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
 import flexjson.JSONSerializer;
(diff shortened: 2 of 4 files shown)


src/main/java/com/labsynch/labseer/chemclasses/indigo/ChemStructureServiceIndigoImpl.java
```diff
@@ -276,7 +277,7 @@ public int[] searchMolStructures(String molfile, StructureType structureType, Se
 
 			return hitList;
 
-		} catch (JpaSystemException sqlException) {
+		} catch (JpaSystemException | PersistenceException sqlException) {
```
brianbolt, line 280: Error we hit in testing was a PersistenceException so I added this extra catch here so that the Bingo error would get propagated up.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -381,159 +382,16 @@ public BulkLoadRegisterSDFResponseDTO registerSdf(BulkLoadRegisterSDFRequestDTO
...
-								logError(e, numRecordsRead, mol, mappings, errorMolExporter, results,
-										errorCSVOutStream);
-							}
-							continue;
-						}
-					} catch (Exception e) {
-						logError(e, numRecordsRead, mol, mappings, errorMolExporter, results, errorCSVOutStream);
-						continue;
+				try{
+					Boolean isNewParent = registerMol(mol, bulkLoadFile, mappings, errorCSVOutStream, registeredCSVOutStream, errorMolExporter, registeredMolExporter, allAliasMaps, numRecordsRead, results, validate, chemist, registerRequestDTO, dryRunCompound, startTime);
```
brianbolt, line 386: This was the major refactor to fix our transaction issue.  registerMol is wrapped in a new transaction the "next loop" can catch the saved structure from the first loop. Additionally, if an error is hit during registerMol, it doesn't roll back the whole transactions (which was the case before the un-released 1.13.7.8

src/main/resources/log4j2.xml
```diff
@@ -13,6 +13,11 @@
     <Logger name="org.springframework.web" level="info"/>
     <Logger name="org.apache.catalina.core" level="info"/>
     <Logger name="org.apache.catalina.core.ContainerBase.[Catalina].[localhost]" level="info"/>
+
+    <!-- Leaving debug transactions here but commented as it can be helpful for knowing when begin/commit of transaction is-->
+    <!-- <Logger name="org.hibernate.engine.transaction.internal.TransactionImpl" level="debug"/>
+    <Logger name="org.springframework.transaction" level="debug"/>
+    <Logger name="org.springframework.orm.jpa" level="debug"/> -->
```
brianbolt, line 20: This was super helpful during debugging the creation and closing of transactions so leaving this breakcrumb in.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -593,6 +451,159 @@ public BulkLoadRegisterSDFResponseDTO registerSdf(BulkLoadRegisterSDFRequestDTO
 		}
 	}
 
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public Boolean registerMol(CmpdRegMolecule mol, BulkLoadFile bulkLoadFile,Collection<BulkLoadPropertyMappingDTO> mappings,
```
brianbolt, line 455: It could be argued that I refactor most of the variables in this function up  a level as they don't really need to be passed in here.  I took the short cut for now...

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -381,159 +382,16 @@ public BulkLoadRegisterSDFResponseDTO registerSdf(BulkLoadRegisterSDFRequestDTO
...
-				// Successful registration/validation: update summary and write
-				// registered/validated record
-				if (isNewParent)
-					numNewParentsLoaded++;
-				else
-					numNewLotsOldParentsLoaded++;
-				writeRegisteredMol(numRecordsRead, mol, metalotReturn, mappings, registeredMolExporter,
-						registeredCSVOutStream, isNewParent, results);
-				if (numRecordsRead % 100 == 0) {
-					logger.info("flushing bulk loader session");
```
bffrost, line 526: I think you can also remove the Session on line 287/288 since we're no longer using it, and remove the import

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -196,23 +196,18 @@ private HashMap<String, Integer> saveDryRunStructures(HashMap<String, CmpdRegMol
 	}
 
 	@Override
+	@Transactional
```
bffrost, line 199: Does this parent function need to be Transactional? Just asking based on the pattern that `registerSDF` is not Transactional while the inner function is.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -777,7 +694,101 @@ public int restandardizeParentStructures(List<Long> parentIds)
 			// Update the percentage.
 			previousPercent = percent;
 		}
-		return parentIds.size();
+		return(p);
+	}
+	
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
```
bffrost, line 700: Ok, so this one we are keeping transactions at the batch level.

bffrost (approved):
LGTM, provided both the acasclient and your restandardization tests are passing now.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Yeah, this stuff isn't doing registrations and then searches. It's goal is to update quickly as all the chekcing has already been done. The bulk loader is different since it registeres "on the fly".

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
brianbolt: Thanks, I removed it.

PR #342, opened 2022-07-20
Add service to swap two parent structures.

iamanandkr:
Create the service to swap two parent structures. Swapping of structures is only allowed if the two parent structures are unique or just duplicates of each other than the swapping.

## Description

Recommended workflow to check:
1. ParentSwapStructuresDTO > ParentSwapStructures
2.  APIParentController > swapParentStructures
3. ParentSwapStructureServiceImpl > swapParentStructurs.

## Related Issue

[ACAS-286](https://jira.schrodinger.com/browse/ACAS-286).

## How Has This Been Tested?

See testing section in the acas PR.

## TODO
~I am seeing issues when querying duplicate parents i.e in the cases where duplicate parents do exist(confirmed via the `/validateParent` API). I am not getting the duplicate parents listed in the `swapParentStructures` service.~

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiParentController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiParentController.java
@@ -1,6 +1,9 @@
 package com.labsynch.labseer.api;
 
 import java.util.Collection;
+import java.util.HashMap;
+
+import flexjson.JSONSerializer;
 
 import com.labsynch.labseer.domain.Parent;
 import com.labsynch.labseer.dto.CodeTableDTO;
@@ -79,23 +82,23 @@ public class ApiParentController {
 		HttpHeaders headers = new HttpHeaders();
 		ParentSwapStructuresDTO parentSwapStructuresDTO = ParentSwapStructuresDTO.fromJSONToParentSwapStructuresDTO(json);
 
-		String msg = String.format(
-			"corpName1=%s & corpName2=%s swap: ",
-			parentSwapStructuresDTO.getCorpName1(),
-			parentSwapStructuresDTO.getCorpName2());
+		String errorMessage;
 		try {
-			if (parentSwapStructuresService.swapParentStructures(parentSwapStructuresDTO)) {
-				logger.info(msg + "success");
-				return new ResponseEntity<String>(headers, HttpStatus.OK);
-			} else {
-				logger.info(msg + "failure");
-				return new ResponseEntity<String>(headers, HttpStatus.INTERNAL_SERVER_ERROR);
-			}
+			errorMessage = parentSwapStructuresService.swapParentStructures(parentSwapStructuresDTO);
 		} catch (Exception e) {
-			logger.error("Caught error while trying to swap parent structures", e);
+			logger.error("Caught error trying to swap parent structures", e);
 			return new ResponseEntity<String>(headers, HttpStatus.INTERNAL_SERVER_ERROR);
 		}
-	}
+
+		boolean hasError = !errorMessage.isEmpty();
+		HashMap<String,Object> response = new HashMap<>();
+		response.put("hasError", hasError);
+		response.put("errorMessage", errorMessage);
+		HttpStatus status = (hasError)? HttpStatus.INTERNAL_SERVER_ERROR : HttpStatus.OK;
+		
+		headers.add("Content-Type", "application/json; charset=utf-8");
+		return new ResponseEntity<String>(new JSONSerializer().serialize(response), headers, status);
+}
 
 	@RequestMapping(value = "/updateParent/metadata/jsonArray", method = RequestMethod.POST, headers = "Accept=application/json")
 	@ResponseBody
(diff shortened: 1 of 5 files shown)


src/main/java/com/labsynch/labseer/api/ApiParentController.java
```diff
@@ -67,6 +74,31 @@ public ResponseEntity<String> updateParent(@RequestBody String json) {
...
+			"corpName1=%s & corpName2=%s swap: ",
+			parentSwapStructuresDTO.getCorpName1(),
+			parentSwapStructuresDTO.getCorpName2());
+		try {
+			if (parentSwapStructuresService.swapParentStructures(parentSwapStructuresDTO)) {
+				logger.info(msg + "success");
+				return new ResponseEntity<String>(headers, HttpStatus.OK);
+			} else {
+				logger.info(msg + "failure");
+				return new ResponseEntity<String>(headers, HttpStatus.INTERNAL_SERVER_ERROR);
```
dalejerikson, line 94: Everything looks great and clean so far! I would suggest adding a JSON message in the `ResponseEntity` for both a successful swap or (even better) an error

src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+    Logger logger = LoggerFactory.getLogger(ParentSwapStructuresServiceImpl.class);
+
+    @Autowired
+    public ParentService parentService;
+
+    @Override
+    public boolean swapParentStructures(ParentSwapStructuresDTO parentSwapStructuresDTO) {
+        String corpName1 = parentSwapStructuresDTO.getCorpName1();
+		String corpName2 = parentSwapStructuresDTO.getCorpName2();
```
dalejerikson, line 25: Fix Indentation

src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+		} catch (Exception e) {
+			logger.error("Caught error while fetching parent", e);
+            return false;
+		}
+
+		ParentValidationDTO validationDTO1;
+		ParentValidationDTO validationDTO2;
+		try {
+			validationDTO1 = parentService.validateUniqueParent(parent1);
+			validationDTO2 = parentService.validateUniqueParent(parent2);
```
dalejerikson, line 41: I lack knowledge of this but is `validateUniqueParent` needed if it was instantiated by `findParentsByCorpNameEquals` above

dalejerikson:
Nice clean code! Easy to read and self explanatory. Just have one question and a few minor suggested changes.

dalejerikson (dismissed):

iamanandkr:
Please hold off reviewing the PR. I need to fix things in here.

iamanandkr:
> Please hold off reviewing the PR. I need to fix things in here.

PR is good to be reviewed for the code logic.

src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
iamanandkr: If the item exists in the db then it should already be valid and here the logic is merely sanity check. I am open to removing the check from here entirely.

src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+		for (int parentCdId : parentCdIds) {
+			for (Parent parent : Parent.findParentsByCdId(parentCdId).getResultList()) {
+				String parentStereoCategory = parent.getStereoCategory().getCode();
+				if (!parentStereoCategory.equalsIgnoreCase(stereoCategory)) {
+					continue;
+				}
+				String parentStereoComment = parent.getStereoComment();
+				if (parentStereoComment == null && stereoComment == null) {
+					matchingParents.add(parent);
+				} else if (parentStereoComment == null || stereoComment == null) {
```
bffrost, line 104: I think there's a possible complication when one is null and the other is empty string. Users don't have a way to truly null out a stereo comment, so we should consider null and empty string as equal to each other.
We handle this case elsewhere, although not that elegantly: https://github.com/mcneilco/acas-roo-server/blob/release/2022.1.x/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java#L806

bffrost (changes requested):
Spotted a small unhandled edge case. Otherwise looks good to me!

src/main/java/com/labsynch/labseer/api/ApiParentController.java
iamanandkr: Done. Let me know if the output response looks good or should I add/remove any fields. cc'ing @bffrost too!

src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
iamanandkr: I am keeping the checks for now!

PR #344, opened 2022-07-27
Reparent lot route updates ACAS-216-reparent-lot

brianbolt:
## Description
Updated reparent lot routes to 
 - return a structured output which contains
    -  original parent corp name
    -  original lot corp name
    - original lot number
    - the new lot as updated (or simulated via new dry run flag) 
 - update experimental data
    - renames codeType "codeValue" codeValue "batch code" values on subject, treatment, experiment and protocol levels
 - deletes old parent if it is orphaned
 - optional dry run mode which returns the expected output lot (including updated corp name and lot number)

## Related Issue
ACAS-216

## How Has This Been Tested?
Ran acas client tests and tests described on acas node layer ACAS-216 pull request

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
@@ -129,13 +129,14 @@ public class ApiParentLotController {
 
 	@RequestMapping(value = "/reparentLot", method = RequestMethod.POST, headers = "Accept=application/json")
 	@ResponseBody
-	public ResponseEntity<String> reparentLot(@RequestBody String json) {
+	public ResponseEntity<String> reparentLot(@RequestBody String json, @RequestParam(value = "dryRun", required = false, defaultValue = "true") Boolean dryRun) {
 		HttpHeaders headers = new HttpHeaders();
 		headers.add("Content-Type", "application/json; charset=utf-8");
 		try {
 			ReparentLotDTO lotDTO = ReparentLotDTO.fromJsonToReparentLotDTO(json);
 			ReparentLotResponseDTO reparentLotDTO = lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(),
-					lotDTO.getModifiedBy(), true, true);
+				lotDTO.getModifiedBy(), true, true, dryRun);
+
 			return new ResponseEntity<String>(reparentLotDTO.toJson(), headers, HttpStatus.OK);
 		} catch (DupeLotException e) {
 			logger.error("Error saving lot with duplicate name", e);
@@ -149,14 +150,14 @@ public class ApiParentLotController {
 	@Transactional
 	@RequestMapping(value = "/reparentLot/jsonArray", method = RequestMethod.POST, headers = "Accept=application/json")
 	@ResponseBody
-	public ResponseEntity<String> reparentLotArray(@RequestBody String json) {
+	public ResponseEntity<String> reparentLotArray(@RequestBody String json, @RequestParam(value = "dryRun", required = false, defaultValue = "true") Boolean dryRun) {
 		HttpHeaders headers = new HttpHeaders();
 		headers.add("Content-Type", "application/json; charset=utf-8");
 		int lotCount = 0;
 		try {
 			Collection<ReparentLotDTO> lotDTOs = ReparentLotDTO.fromJsonArrayToReparentLoes(json);
 			for (ReparentLotDTO lotDTO : lotDTOs) {
-				lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(), lotDTO.getModifiedBy(), true, true);
+				lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(), lotDTO.getModifiedBy(), true, true, dryRun);
 				lotCount++;
 			}
 			return new ResponseEntity<String>("number of lots reparented: " + lotCount, headers, HttpStatus.OK);
--- a/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
@@ -113,6 +113,7 @@ public class ApiParentLotController {
 		}
 	}
 
+	@Transactional
 	@RequestMapping(value = "/updateLot/metadata/jsonArray", method = RequestMethod.POST, headers = "Accept=application/json")
 	@ResponseBody
 	public ResponseEntity<String> updateLotMetaArray(@RequestBody String json) {
--- a/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
@@ -9,6 +9,7 @@ import com.labsynch.labseer.dto.LotDTO;
 import com.labsynch.labseer.dto.LotsByProjectDTO;
 import com.labsynch.labseer.dto.ParentLotCodeDTO;
 import com.labsynch.labseer.dto.ReparentLotDTO;
+import com.labsynch.labseer.dto.ReparentLotResponseDTO;
 import com.labsynch.labseer.service.LotService;
 import com.labsynch.labseer.service.ParentLotService;
 
@@ -126,7 +127,6 @@ public class ApiParentLotController {
 		}
 	}
 
-	@Transactional
 	@RequestMapping(value = "/reparentLot", method = RequestMethod.POST, headers = "Accept=application/json")
 	@ResponseBody
 	public ResponseEntity<String> reparentLot(@RequestBody String json) {
@@ -134,9 +134,9 @@ public class ApiParentLotController {
 		headers.add("Content-Type", "application/json; charset=utf-8");
 		try {
 			ReparentLotDTO lotDTO = ReparentLotDTO.fromJsonToReparentLotDTO(json);
-			Lot lot = lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(),
-					lotDTO.getModifiedBy());
-			return new ResponseEntity<String>(lot.toJsonIncludeAliases(), headers, HttpStatus.OK);
+			ReparentLotResponseDTO reparentLotDTO = lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(),
+					lotDTO.getModifiedBy(), true, true);
+			return new ResponseEntity<String>(reparentLotDTO.toJson(), headers, HttpStatus.OK);
 		} catch (Exception e) {
 			logger.error("Caught exception updating lot metadata", e);
 			return new ResponseEntity<String>(headers, HttpStatus.NOT_FOUND);
@@ -153,7 +153,7 @@ public class ApiParentLotController {
 		try {
 			Collection<ReparentLotDTO> lotDTOs = ReparentLotDTO.fromJsonArrayToReparentLoes(json);
 			for (ReparentLotDTO lotDTO : lotDTOs) {
-				lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(), lotDTO.getModifiedBy());
+				lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(), lotDTO.getModifiedBy(), true, true);
 				lotCount++;
 			}
 			return new ResponseEntity<String>("number of lots reparented: " + lotCount, headers, HttpStatus.OK);
--- a/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
@@ -10,6 +10,7 @@ import com.labsynch.labseer.dto.LotsByProjectDTO;
 import com.labsynch.labseer.dto.ParentLotCodeDTO;
 import com.labsynch.labseer.dto.ReparentLotDTO;
 import com.labsynch.labseer.dto.ReparentLotResponseDTO;
+import com.labsynch.labseer.exceptions.DupeLotException;
 import com.labsynch.labseer.service.LotService;
 import com.labsynch.labseer.service.ParentLotService;
 
@@ -112,7 +113,6 @@ public class ApiParentLotController {
 		}
 	}
 
-	@Transactional
 	@RequestMapping(value = "/updateLot/metadata/jsonArray", method = RequestMethod.POST, headers = "Accept=application/json")
 	@ResponseBody
 	public ResponseEntity<String> updateLotMetaArray(@RequestBody String json) {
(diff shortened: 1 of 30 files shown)


src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
```diff
@@ -111,7 +113,6 @@ public ResponseEntity<String> updateLotMetadata(@RequestBody String json) {
 		}
 	}
 
-	@Transactional
```
brianbolt, line 114: Transactional was not necessary at this level

src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
```diff
@@ -126,34 +127,37 @@ public ResponseEntity<String> updateLotMetaArray(@RequestBody String json) {
...
```
brianbolt, line 129: Transactional was not necessary at this level

src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
```diff
@@ -126,34 +127,37 @@ public ResponseEntity<String> updateLotMetaArray(@RequestBody String json) {
...
+	public ResponseEntity<String> reparentLot(@RequestBody String json, @RequestParam(value = "dryRun", required = false, defaultValue = "true") Boolean dryRun) {
 		HttpHeaders headers = new HttpHeaders();
 		headers.add("Content-Type", "application/json; charset=utf-8");
 		try {
 			ReparentLotDTO lotDTO = ReparentLotDTO.fromJsonToReparentLotDTO(json);
-			Lot lot = lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(),
-					lotDTO.getModifiedBy());
-			return new ResponseEntity<String>(lot.toJsonIncludeAliases(), headers, HttpStatus.OK);
+			ReparentLotResponseDTO reparentLotDTO = lotService.reparentLot(lotDTO.getLotCorpName(), lotDTO.getParentCorpName(),
+				lotDTO.getModifiedBy(), true, true, dryRun);
```
brianbolt, line 138: The hard coded boolean are for updating lot inventory and containers which we always want to do with this route.  I imaged other scenarios where we wouldn't want to do this.  I.e. if there was an admin utility which was reparenting a lot and then replacing the original lot so it wanted to leave the assay and inventory data inlace.

src/main/java/com/labsynch/labseer/domain/AnalysisGroupValue.java
```diff
@@ -1970,4 +1970,33 @@ public void setLsState(AnalysisGroupState lsState) {
 	public AnalysisGroupValue() {
 		super();
 	}
+
+    public static TypedQuery<AnalysisGroupValue> findAnalysisGroupValuesByLsKindEqualsAndCodeValueEquals(
```
brianbolt, line 1974: Didn't want to spend the time to try and abstract this query so you'll see this repeated on many different levels (subject, treatment, experiment..etc.)

src/main/java/com/labsynch/labseer/exceptions/DupeLotException.java
```diff
@@ -1,6 +1,6 @@
 package com.labsynch.labseer.exceptions;
 
-public class DupeLotException extends Exception {
+public class DupeLotException extends RuntimeException {
```
brianbolt, line 3: This was necessary to make sure the transaction failed.  Without a runtime exception, transactions don't roll back on thrown errors.

src/main/java/com/labsynch/labseer/service/AnalysisGroupServiceImpl.java
```diff
@@ -540,4 +542,36 @@ private AnalysisGroup getOrCreateAnalysisGroup(
 		return analysisGroup;
 	}
 
+	@Override
+	@Transactional
+	public int renameBatchCode(String oldCode, String newCode, String modifiedByUser, Long transactionId) {
```
brianbolt, line 547: Another repeated function you'll see in this review on multiple levels, I didn't abstract it out.

src/main/java/com/labsynch/labseer/service/AssayServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+		lsTransaction.setRecordedDate(new Date());
+		lsTransaction.setRecordedBy(modifiedByUser);
+		lsTransaction.setComments("Rename batch code " + oldCode + " to " + newCode);
+		lsTransaction.persist();
+
+		int protocolValuesAffected = protocolService.renameBatchCode(oldCode, newCode, modifiedByUser, lsTransaction.getId());
+		int experimentValuesAffected = experimentService.renameBatchCode(oldCode, newCode, modifiedByUser, lsTransaction.getId());
+		int analysisGroupValuesAffected = analysisGroupService.renameBatchCode(oldCode, newCode, modifiedByUser, lsTransaction.getId());
+		int treatmentGroupValuesAffected = treatmentGroupService.renameBatchCode(oldCode, newCode, modifiedByUser, lsTransaction.getId());
+		int subjectValuesAffected = subjectService.renameBatchCode(oldCode, newCode, modifiedByUser, lsTransaction.getId());
```
brianbolt, line 45: There was never really a need to return this data back.  On return from this function, I just call get lot dependencies on the new lot to show the results of the change.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -80,52 +87,107 @@ public Lot updateLotWeight(Lot lot) {
...
+
+		Lot queryLot = Lot.findLotsByCorpNameEquals(lotCorpName).getSingleResult();
+		reparentLotResponseDTO.setOriginalLotNumber(queryLot.getLotNumber());
 
+		SaltForm saltForm = queryLot.getSaltForm();
 		Parent adoptiveParent = Parent.findParentsByCorpNameEquals(parentCorpName).getSingleResult();
+		Set isoSalts = saltForm.getIsoSalts();
+		isoSalts.size();
+		if(dryRun) {
+			Lot.entityManager().detach(queryLot);
```
brianbolt, line 119: If dry run, we need to detach the objects so their data doesn't get persisted automatically by hibernate.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -80,52 +87,107 @@ public Lot updateLotWeight(Lot lot) {
...
+
+		Boolean deleteOriginalParentAfterReparent = false;
+		Long originalParentId = queryLot.getParent().getId();
+		if (lotOrphanLevel.equals(LotOrphanLevel.Parent)) {
+			deleteOriginalParentAfterReparent = true;
+		}
+		
 		// renumber lot -- just increment to next number
-		Integer newLotNumber = Lot.getMaxParentLotNumber(adoptiveParent) + 1;
+		Integer newLotNumber = predictReparentLotNumber(adoptiveParent, queryLot);
```
brianbolt, line 133: Added new function to account for max auto lot property to maintain the same lot number on reparenting.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -80,52 +87,107 @@ public Lot updateLotWeight(Lot lot) {
...
+		reparentLotResponseDTO.setNewLot(queryLot);
+
+		if(deleteOriginalParentAfterReparent) {
+			reparentLotResponseDTO.setOriginalParentDeleted(true);
+		} else {
+			reparentLotResponseDTO.setOriginalParentDeleted(false);
+		}
+
+		if(!dryRun) {
+			saltForm.merge();
```
brianbolt, line 172: wrapped all merges and assay data updates into this if block so we don't actually perform updates.  This means that the dry run doesn't really cover assay data changes but the check dependency route allows us to get the expected changes anyway.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -608,4 +670,39 @@ public void deleteLot(Lot lot) {
 		lot.remove();
 	}
 
+	public enum LotOrphanLevel {
+		Parent, SaltForm, NONE
+	}
+
+	@Override
+	public LotOrphanLevel checkLotOrphanLevel(Lot lot) {
```
brianbolt, line 678: New method  for reuse with delete lot to check if the salt form, parent, or none would be orphaned by the lot going away (reparented or deleted)

src/main/java/com/labsynch/labseer/service/ParentLotServiceImpl.java
```diff
@@ -123,31 +124,14 @@ public void updateLot() {
 	@Override
 	public void deleteLot(Lot lot) {
 
-		Parent parent = Parent.findParent(lot.getParent().getId());
```
brianbolt, line 126: All these changes in this file are about using the new `lotService.checkLotOrphanLevel` method

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -80,52 +87,107 @@ public Lot updateLotWeight(Lot lot) {
...
+
+		if(deleteOriginalParentAfterReparent) {
+			reparentLotResponseDTO.setOriginalParentDeleted(true);
+		} else {
+			reparentLotResponseDTO.setOriginalParentDeleted(false);
+		}
+
+		if(!dryRun) {
+			saltForm.merge();
+			saltFormService.updateSaltWeight(saltForm);
```
bffrost, line 173: It's strange to me that this is called _after_ you `merge` the saltForm. Have you double checked that the updated Salt Weight actually goes through properly?

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -80,52 +87,107 @@ public Lot updateLotWeight(Lot lot) {
...
+			// recalculate salt form weight
+			queryLot.merge();
+			if(deleteOriginalParentAfterReparent) {
+				Parent originalParent = Parent.findParent(originalParentId);
+				Parent.entityManager().refresh(originalParent);
```
bffrost, line 178: Worth a code comment on why you're doing this `refresh`. I only understand because you and I talked about it. It's non-obvious, so a comment would be a great way to record that knowledge.

bffrost (changes requested):
This is really well structured and easy to follow. Nice work @brianbolt .
I've added a handful of questions and requests for code comments on parts of the code that are particularly tricky, but this is in great shape and should be ready to merge after you address those.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -80,52 +87,107 @@ public Lot updateLotWeight(Lot lot) {
...
+		reparentLotResponseDTO.setOriginalLotCorpName(lotCorpName);
+		reparentLotResponseDTO.setModifiedBy(modifiedByUser);
+
+		Lot queryLot = Lot.findLotsByCorpNameEquals(lotCorpName).getSingleResult();
+		reparentLotResponseDTO.setOriginalLotNumber(queryLot.getLotNumber());
 
+		SaltForm saltForm = queryLot.getSaltForm();
 		Parent adoptiveParent = Parent.findParentsByCorpNameEquals(parentCorpName).getSingleResult();
+		Set isoSalts = saltForm.getIsoSalts();
+		isoSalts.size();
```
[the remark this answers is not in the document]
brianbolt: Thanks, I added some notes in the code about this. It's a little strange but you have to essentially loop through the items, or call "size" to actually initialize and use these items.  After this, I am detaching salt form and later trying to use iso salts (which haven't been fetched) to generate the salt form corp name.  Calling size was the easiest way I saw to initialize the item.  I moved this to only run if within the dry run if block to make it more obvious along with comments.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
brianbolt: I added some comments now.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
brianbolt: Thanks, fixed this. Would have been a bug too if we used iso salts.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
brianbolt: Thanks added some notes to the best of my knowledge on why this was necessary.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
```diff
@@ -80,52 +87,107 @@ public Lot updateLotWeight(Lot lot) {
...
+		if(dryRun) {
+			Lot.entityManager().detach(queryLot);
+			SaltForm.entityManager().detach(saltForm);
+			Parent.entityManager().detach(adoptiveParent);
+		}
+
+		LotOrphanLevel lotOrphanLevel = lotService.checkLotOrphanLevel(queryLot);
```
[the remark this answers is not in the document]
brianbolt: Added some comments.

src/main/java/com/labsynch/labseer/api/ApiParentLotController.java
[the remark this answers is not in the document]
brianbolt: good catch I did NOT mean to remove this. just added it back

src/main/java/com/labsynch/labseer/service/AnalysisGroupServiceImpl.java
```diff
@@ -540,4 +542,36 @@ private AnalysisGroup getOrCreateAnalysisGroup(
...
+
+		List<AnalysisGroupValue> values = valueQuery.getResultList();
+		logger.info("Found " + values.size() + " analysis group values with batch code " + oldCode);
+		// Loop through each code value, ignore it and create a new one with the new code
+		for (AnalysisGroupValue value : values) {
+			value.setIgnored(true);
+			value.setModifiedBy(modifiedByUser);
+			value.setModifiedDate(new Date());
+			value.setLsTransaction(transactionId);
+			value.persist();
```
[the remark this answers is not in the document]
brianbolt: I tried removing both persists (existing value and new value) and the existing value was updated (ignored) and the new value was not saved.
I removed just the persist on the original value but kept permits on the new value and that worked.  Existing value was updated (ignored) and new value was saved.

I just checked that change in.

src/main/java/com/labsynch/labseer/service/LotServiceImpl.java
brianbolt: I didn't rerun the test after my update to fix this and found that tests were failing. Looking into it.

brianbolt: I found that `saltFormService.updateSaltWeight` was trying to call `saltForm.merge()`. I removed that call in ` `saltFormService.updateSaltWeight` and added it to the only time it was being used in LotServiceImpl.java on the `updateLotWeight` method.

brianbolt: Tests now pass again

bffrost (approved):
LGTM!

PR #345, opened 2022-07-28
DE - Salt Browser

dalejerikson:
This PR represents the first version of the backend routes required for the ACAS Salt Browser to be functional. 

ACAS PR: https://github.com/mcneilco/acas/pull/967

**Status of Each Feature and Sub-Feature**

Query Route 

- Added query to return any salts in the database where the query string matches (LIKE) to any abbrev or names of salts in the database 

Create Salt Route 

- Able to create and register salts successfully 
- Requires JSON salt data which contains a name, abbrev, and mol structure
- Dryrun mode calculates molecular properties but does not save any Salt 
- Validation is done on structure, name, and abbreviation fields being filled 

Edit Salt Route

- Able to change the structure, name, and/or abbrev of any Salt 
- Has a dryrun mode that will return list of warnings/errors including cascading changes and affected dependencies 
- Edit salt route calls restandardization function as well as a dependency check 

Delete Salt Route

- Route uses id of Salt to find; If it exists and there are no lot dependencies then the salt can be deleted 
- Route to delete salt includes validation check to determine if salt is present in any lots; if it is then an error is returned 

Get SDF Route 

- returns SDF string of all salts within database 
- relevant salt properties are included as SDF properties


src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+				List<Salt> saltsByAbbrev = Salt.findSaltsByAbbrevEquals(newSalt.getAbbrev()).getResultList();
+				if (saltsByAbbrev.size() > 1) {
+					logger.error("Number of salts found: " + saltsByAbbrev.size());
+					validSalt = false;
+					ErrorMessage error = new ErrorMessage();
+					error.setLevel("error");
+					error.setMessage("Duplicate salt abbreviation. Another salt exist with the same abbreviation.");
+					warnings.add(error);
+				}
+				List<Salt> saltsByFormula = Salt.findSaltsByFormulaEquals(newSalt.getFormula()).getResultList();
```
dalejerikson, line 171: Now that I look at this again there is such a thing as isomers so I will want to change this to checking duplicate structures

dalejerikson: maybe I will keep this but set the level to warning

src/main/java/com/labsynch/labseer/service/SaltStructureServiceImpl.java
```diff
@@ -65,27 +65,106 @@ public Salt saveStructure(Salt salt) throws CmpdRegMolFormatException {
...
-			return Salt.findSalt(salt.getId());
-		} else {
-			return null;
+			if (updated) {
+				return Salt.findSalt(salt.getId());
+			} else {
+				return null;
+			}
 		}
+		catch (Exception e)
```
dalejerikson, line 90: Might want to make this catch only CmpdRegMolFormatException

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+					warning.setLevel("warning");
+					warning.setMessage("The abbreviation of this salt will be changed.");
+					warnings.add(warning);
+				}
+				if(!newSalt.getName().equals(oldSalt.getName())){
+					ErrorMessage warning = new ErrorMessage();
+					warning.setLevel("warning");
+					warning.setMessage("The name of this salt will be changed.");
+					warnings.add(warning);
+				}
```
dalejerikson, line 138: To trim or not to trim...hmmmm.....

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+
+				newSalt.setAbbrev(newSalt.getAbbrev().trim());
+				newSalt.setName(newSalt.getName().trim());
+				newSalt.setFormula(saltStructureService.calculateFormula(newSalt));
+
+				// This doesn't always work; Might need to convert to mol and then compare
```
dalejerikson, line 144: I will want to implement a CompareTo method of some sort for salt structs

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
bffrost: I like the idea of keeping it but with a warning.

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+				if(!newSalt.getName().equals(oldSalt.getName())){
+					ErrorMessage warning = new ErrorMessage();
+					warning.setLevel("warning");
+					warning.setMessage("The name of this salt will be changed.");
+					warnings.add(warning);
+				}
+
+				newSalt.setAbbrev(newSalt.getAbbrev().trim());
+				newSalt.setName(newSalt.getName().trim());
+				newSalt.setFormula(saltStructureService.calculateFormula(newSalt));
```
bffrost, line 142: Should we be calculating the molecular weight and charge at this point too?

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
bffrost: Can we just calculate weight & charge, and skip comparing structures? Structure comparison is a rabbit hole, and we should avoid it if we can.

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+					error.setMessage("Duplicate salt formula. Another salt exist with the same formula.");
+					warnings.add(error);
+				}
+				try { 
+					PurgeSaltDependencyCheckResponseDTO dependencyReport  = saltService.checkDependentData(oldSalt);
+					boolean hasDependencies = ! dependencyReport.isCanPurge();
+					if(hasDependencies){
+						ErrorMessage warning = new ErrorMessage();
+						warning.setLevel("warning");
+						warning.setMessage("This salt has experiment/lot dependencies. An attempt will be made to restandardize.");
```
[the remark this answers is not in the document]
bffrost: I answered my own question later in the code review. I did not remember / know about that `standardizationService.restandardizeLots` function which recalculates lot molecular weight.

I'm not quite sure what that has to do with dependent data though. Could you elaborate on the thinking here?

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+						error.setMessage("Error updating salt: " + e.getMessage());
+						warnings.add(error);
+
+						return new ResponseEntity<String>(ErrorMessage.toJsonArray(warnings), headers, HttpStatus.BAD_REQUEST);
+					}
+				}
+				else if (!validSalt)
+				{
+					return new ResponseEntity<String>(ErrorMessage.toJsonArray(warnings), headers,
+					HttpStatus.EXPECTATION_FAILED);
```
bffrost, line 231: I don't think this is the right status code. This seems to have something to do with an "Expect" header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417

I think a BAD_REQUEST error would be more appropriate here

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+		if (salt == null) {
+			return new ResponseEntity<String>(headers, HttpStatus.NOT_FOUND);
+		}
+		else
+		{
+			// Need to Do Validation Of No Lot Association Before Final Removal
+			boolean lotDependency = false; 
+			PurgeSaltDependencyCheckResponseDTO dependencyReport  = saltService.checkDependentData(salt);
+			lotDependency = ! dependencyReport.isCanPurge();
+			// Need to Get Dependency Bool From Report JSON String
```
bffrost, line 268: Not sure how to interpret this comment

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -95,21 +293,63 @@ public ResponseEntity<String> listJson() {
...
+		headers.setExpires(0); // Expire the cache
+
+		Salt salt = null; 
+		boolean dryrun = true; 
+		try
+		{
+			ObjectMapper objectMapper = new ObjectMapper();
+
+			JsonNode rootNode = objectMapper.readTree(json);
+			JsonNode dryRunNode = rootNode.path("dryrun");
```
bffrost, line 327: Similar thing here - I think it'd make sense to put the `dryrun` as a RequestParameter

src/main/java/com/labsynch/labseer/service/StandardizationService.java
```diff
@@ -29,6 +29,8 @@ String getStandardizationDryRunReportFiles(StandardizationDryRunSearchDTO search
 	String executeStandardization(String username, String reason)
 			throws IOException, CmpdRegMolFormatException, StandardizerException;
 
+	public int restandardizeLots(List<Long> parentIds);
```
bffrost, line 32: How strange that this function existed but wasn't exposed. Just a nit - the term "restandardize" now has a specific connotation of running chemical structures through a "standardizer" or "preprocessor" or "structure cleaner", which isn't what this function does. It'd be nice to name this something more obvious, like "recalculateLotMolWeights"

src/main/java/com/labsynch/labseer/service/SaltStructureServiceImpl.java
bffrost: I agree

src/main/java/com/labsynch/labseer/service/SaltStructureServiceImpl.java
```diff
@@ -65,27 +65,106 @@ public Salt saveStructure(Salt salt) throws CmpdRegMolFormatException {
...
 	}
 
+    public Salt edit(Salt oldSalt, Salt newSalt){
+        try
+        {
+            oldSalt.setAbbrev(newSalt.getAbbrev());
+            oldSalt.setName(newSalt.getName());
+
+            oldSalt.setOriginalStructure(newSalt.getMolStructure());
+        	oldSalt.setMolStructure(newSalt.getMolStructure());
```
bffrost, line 105: nit: indentation is a little wonky in this file

src/main/java/com/labsynch/labseer/service/SaltStructureServiceImpl.java
```diff
@@ -65,27 +65,106 @@ public Salt saveStructure(Salt salt) throws CmpdRegMolFormatException {
...
+			return null;
+        }
+    }
+
+	public double calculateWeight(Salt salt) {
+        double weight = -1.0;
+        try
+        {
+            CmpdRegMolecule mol = chemStructureService.toMolecule(salt.getMolStructure());
+		    weight = mol.getExactMass();
```
bffrost, line 127: We should be careful here. There are two methods to calculate molecular weight: `getMass` which calculates average molecular weight based on isotopic abundance and `getExactMass` which calculates the monoisotopic mass. We're using `getMass` above, so I think this line should use `mol.getMass();` too. `getMass` is sortof our default, then in specific scenarios we may use `getExactMass` when the monisotopic mass is specifically requested.

src/main/java/com/labsynch/labseer/service/SaltStructureServiceImpl.java
```diff
@@ -65,27 +65,106 @@ public Salt saveStructure(Salt salt) throws CmpdRegMolFormatException {
...
+            return weight;
+        }
+        catch (Exception e)
```
bffrost, line 130: I'm not sure why these general "catch" statements are here. Is there a reason we don't simply let any errors get thrown and bubble up to whatever code is calling `calculateWeight`? I'm generally opposed to "swallowing" errors unless we have a specific reason for a specific type of error.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +147,207 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+			sdfWriter.close();
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+		logger.debug("ou	tput SDF: " + sdfWriter.getBufferString());
+		return sdfWriter.getBufferString();
+	}
+
+	// Helper Method Used to Check Dependencies Within ACAS
+	private Map<String, HashSet<String>> checkACASDependencies(
```
bffrost, line 190: Brian Bolt did some work recently to refactor and simplify dependency checking. You should have a look at his code and chat with him, and see if there are places where you can reuse functions he's written, or follow that same pattern.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +147,207 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+	// Method to Check for Dependent Data 
+	public PurgeSaltDependencyCheckResponseDTO checkDependentData(Salt salt)
+	{
+		Map<String, HashSet<String>> acasDependencies = new HashMap<String, HashSet<String>>();
+		Map<String, HashSet<String>> cmpdRegDependencies = new HashMap<String, HashSet<String>>();
+		HashSet<String> dependentSingleRegLots = new HashSet<String>();
+		int numberOfParents = 0;
+		int numberOfSaltForms = 0;
+		int numberOfLots = 0;
+		Collection<Parent> parents = Parent.findParentsByCdId(salt.getCdId()).getResultList();
```
bffrost, line 233: This line seems to be looking up Parent compounds whose structure ID matches the structure ID of the salt. This isn't correct.

To find dependencies for Salts, we'll need to find any SaltForms that contain this Salt, then find the dependencies of those SaltForms. We should chat about this.

bffrost (changes requested):
This is a good start to the backend logic. I think it'll help clean up a bunch if you study some of Brian's recent dependency checking code, and then try to reuse as much as that as you can here. Please comment back on any of my feedback that doesn't make sense, and we can work on this also during pair programming.

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
dalejerikson: So, just to confirm, there should be no alert if someone makes an isomer of a salt?

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -175,9 +422,19 @@ public ResponseEntity<String> updateFromJson(@RequestBody String json) {
 			if (updatedSalt == null) {
 				return new ResponseEntity<String>(ErrorMessage.toJsonArray(errors), headers, HttpStatus.CONFLICT);
 			} else {
-				return new ResponseEntity<String>(updatedSalt.toJson(), headers, HttpStatus.OK);
+				try
+				{
+					String saltMol = updatedSalt.toJson();
```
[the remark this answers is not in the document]
dalejerikson: I don't believe I use `updateFromJSON`. However, I just wanted to keep the ApiSaltController consistent with the updates I was making. I'm confused on the context of what you're saying here.

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
dalejerikson: Yes, I will do this since for the weight and charge warning checks

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
dalejerikson: Will replace this.

src/main/java/com/labsynch/labseer/service/SaltStructureServiceImpl.java
dalejerikson: Will remove them or make more specific catches

src/main/java/com/labsynch/labseer/service/SaltStructureServiceImpl.java
dalejerikson: Gotcha. Thanks for the background. Didn't know that :). Will replace this.

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+		headers.add("Cache-Control", "no-store, no-cache, must-revalidate"); // HTTP 1.1
+		headers.add("Pragma", "no-cache"); // HTTP 1.0
+		headers.setExpires(0); // Expire the cache
+		if (oldSalt == null) {
+			return new ResponseEntity<String>("Salt could not be found", headers, HttpStatus.NOT_FOUND);
+		}
+		else
+		{
+			try {
+				ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
dalejerikson: Yep!

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +95,191 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+
+				JsonNode rootNode = objectMapper.readTree(json);
+				JsonNode dryRunNode = rootNode.path("dryrun");
```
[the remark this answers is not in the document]
dalejerikson: Your thoughts are correct. Will do that. Thanks for the suggestion!

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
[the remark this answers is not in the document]
dalejerikson: The thought was if a lot is dependent on a salt's existence and then the salt is updated then the lot should be restandardized and the weights should be updated. I don't have the best grasp of the object relations in ACAS and how Salts, Salt Forms, Lots, Containers, etc. are all connected and related together. Is there a chart or documentation that could help with this?

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
dalejerikson: This code was based off of code Brian B. messaged me a few weeks ago but I will follow up and see if it is still representative of what I should be doing.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
dalejerikson: Ah makes sense. Happy to chat about this further. I need to gain a better understanding of identifiers and object relationships in ACAS.

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
bffrost: Confirmed. No alert if someone makes an isomer of a salt. We can always add something like that later, but for now let's keep it simple.

dalejerikson:
@bffrost I'd like to review some of the dependency checking and restandardization with you during our scheduled meeting today since the logic is intense and it extends to regions of the system I'm unfamiliar with. Specific cases also include:

~ Checking if the salt abbrev is included in Lot Corp Names
~ If the Lot Corp Name changed, cascade the update to "batch code" values within Analysis Groups and Lot Inventory Containers

dalejerikson:
Rebased Branch

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
dalejerikson: Implemented

dalejerikson:
Backend Ready for Review

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -82,6 +94,110 @@ public ResponseEntity<String> showJson(@PathVariable("id") Long id) {
...
+						oldSalt = saltStructureService.edit(oldSalt, newSalt);
+						try
+							{
+								saltService.updateDependencies(oldSalt);
+								// Return Response of Any Warnings / Errors from Validations / Dependencies Done Prior 
+								return new ResponseEntity<String>(ErrorMessage.toJsonArray(warnings), headers, HttpStatus.OK);
+							}
+						catch (Exception e)
+							{
+								return new ResponseEntity<String>("ERROR: Restandardization Issue:" + e.getMessage(), headers,
```
bffrost, line 130: nit: this term restandardization is going to throw people off. Can you rename this error to "ERROR: Hit an issue updating salt dependencies:"

src/main/java/com/labsynch/labseer/service/SaltService.java
```diff
@@ -1,7 +1,24 @@
...
+import com.labsynch.labseer.dto.PurgeSaltDependencyCheckResponseDTO;
+
+
 public interface SaltService {
 
 	int loadSalts(String saltSD_fileName);
 
+	String exportSalts();
+
+	ArrayList<Long> getAllParentIDs(Salt salt);
```
bffrost, line 15: nit: I think this function is no longer being used, so it can be removed.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+		try {
+			sdfWriter.close();
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+		logger.debug("ou	tput SDF: " + sdfWriter.getBufferString());
+		return sdfWriter.getBufferString();
+	}
+
+	public boolean isValidSaltEdit(Salt oldSalt, Salt newSalt)
```
bffrost, line 211: This function seems redundant with the much more fleshed out `validateSaltEdit` function. I think you're already doing these two checks within `validateSaltEdit`. Perhaps you could run `validateSaltEdit` then check whether it has any messages with level = `error` to derive your boolean "validSalt" ?

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+		for (IsoSalt isoSalt : isoSaltsSet)
+		{
+			SaltForm tempSaltForm = isoSalt.getSaltForm();
+			saltFormSet.add(tempSaltForm);
+		}
+
+		// Getter of Lots from IsoSalt 
+
+		Set<Lot> lotSet = new HashSet<Lot>(); 
+		for (SaltForm saltForm : saltFormSet)
```
bffrost, line 393: nit: you could easily do this with a nested for loop within the `isoSalt` loop you're doing above.
```
Set<Lot> lotSet = new HashSet<Lot>(); 
for (IsoSalt isoSalt : isoSaltsSet){
	SaltForm saltForm = isoSalt.getSaltForm();
	for (SaltForm saltForm : saltFormSet){
		Set<Lot> lots = saltForm.getLots();
		lotSet.addAll(lots);
}
```

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+		{
+			Set<Lot> tempLotSet = saltForm.getLots();
+			lotSet.addAll(tempLotSet);
+		}
+
+		for (Lot lot : lotSet){
+			// Update Lot Corp Name If Format Uses Salt Abbrev
+			if (!propertiesUtilService.getCorpBatchFormat().equalsIgnoreCase("cas_style_format")) {
+				String newCorpName = lotService.generateCorpName(lot);
+				lot.setCorpName(newCorpName);
```
bffrost, line 403: I think you're missing a check whether the newCorpName is different from the old corp name. If it's not different, then we can skip the calls to `assayService.renameBatchCode` and `containerService.renameBatchCode` and the logging of "new lot corp name"

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+                containerService.renameBatchCode(lot.getCorpName(), newCorpName, "SaltService", null);
+
+			}
+			logger.info("new lot corp name: " + lot.getCorpName());
+			// Recalculate Lot Molecular Weights
+			lot.setLotMolWeight(Lot.calculateLotMolWeight(lot));		
+		}
+	}
+
+	public ArrayList<Long> getAllParentIDs(Salt salt)
```
bffrost, line 415: nit: This unused function can be removed to avoid confusion.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+		catch (NullPointerException e)
+		{
+			// Query Result Empty
+			e.printStackTrace();
+			return new ArrayList<Long>();
+		}
+	}
+
+	// Helper Method Used to Check Dependencies Within ACAS
+	private Map<String, HashSet<String>> checkACASDependencies(
```
bffrost, line 465: nit: This unused function can be removed too.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+					}
+				}
+			}
+			return acasDependencies;
+		}
+
+	// Helper Method to Check Dependent Containers; Utilized in Method Below It
+
+	// batchCodes == Lot Corp Name
+	private Collection<ContainerBatchCodeDTO> checkDependentACASContainers(Set<String> batchCodes)
```
bffrost, line 491: nit: This unused function can be removed too.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+		return warnings;
+	}
+
+	public void updateDependencies(Salt salt)
+	{
+		
+
+		// Get IsoSalts From Salt 
+		TypedQuery<IsoSalt> isoSaltQuery = IsoSalt.findIsoSaltsBySalt(salt);
+		Collection<IsoSalt> isoSalts =  isoSaltQuery.getResultList();
```
bffrost, line 378: hypernit: you could do this in a one-liner:
```
Collection<IsoSalt> isoSalts =  IsoSalt.findIsoSaltsBySalt(salt).getResultList();
```
This is typically what we do, and I find it a little easier to follow.

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+			Collection<ContainerBatchCodeDTO> responseDTOs = ContainerBatchCodeDTO
+					.fromJsonArrayToContainerBatchCoes(responseJson);
+			return responseDTOs;
+		}
+
+	// Method to Check for Dependent Data 
+	public PurgeSaltDependencyCheckResponseDTO checkDependentData(Salt salt)
+	{
+		// Get IsoSalts From Salt 
+		TypedQuery<IsoSalt> isoSaltQuery = IsoSalt.findIsoSaltsBySalt(salt);
```
bffrost, line 505: Same hypernit here

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -99,4 +169,369 @@ private void saveSalt(CmpdRegMolecule mol) throws IOException, CmpdRegMolFormatE
...
+		Collection<IsoSalt> isoSalts =  isoSaltQuery.getResultList(); 
+		Set<IsoSalt> isoSaltsSet = new HashSet<IsoSalt>(isoSalts);
+
+		// Getter of IsoSalt to SaltForm 
+
+		Set<SaltForm> saltFormSet = new HashSet<SaltForm>();
```
bffrost, line 511: same nit here about combining the for loops

bffrost: nit: If you really wanted to DRY this up, you might consider a small helper function: `findLotsBySalt` that takes this bit of repeated logic out of both `checkDependentData` and `updateDependencies`

bffrost (changes requested):
@dalejerikson This is getting really close. By and large, the logic all looks right to me. I've noted some leftover / unused functions that you should clean up, and a few other nits, but this will be good to go once you resolve those. Nice work!

bffrost (approved):
LGTM!!

PR #346, opened 2022-07-29
set timeout on calls to processor for bbchem cmpd reg implmentation ACAS-370

brianbolt:
## Description
ACAS is not properly setting timeout on calls to processor, this fix gets the timeout from the configs and if not set defaults to 900 seconds.

## Related Issue
ACAS-370

## How Has This Been Tested?
Ran test and verified that bbchem printed the timeout:

```
2022-07-29 17:26:22 INFO [pid-133] bbchem.preprocessor.views Starting to process 1 structures with a timeout of 900s...
```


bffrost (approved):
LGTM!

PR #348, opened 2022-08-08
Catch no original structure found when running dry run standardization

brianbolt:
## Description
The old version of reparent lot did not catch when it was leaving behind parents with no salt_forms.  This has been fixed but we have systems which have this data state.  This fix allows standardization dry run and standardization to run successfully in this state.  It looks like we already had a check for if we were returning null but we weren't catching a no result exception.

## Related Issue
ACAS-388

## How Has This Been Tested?
Ran the code on a system with the required error state (Parents with no salt forms).


bffrost (approved):
LGTM. Please proceed with merging this into 2022.1.x so we can include this in the upcoming release.

PR #349, opened 2022-08-10
Upgraded JChem to look at Repository for JAR

philippcheung:
Upgraded JChem to use JChem private repo -- still need to figure out the authentication part of the deployment because it involves putting API Keys into a file outside of the repo.

## Description
- Made changes to the POM to upgrade to tomcat JDK 11 image and use Java 11 in related dependencies
- Made changes to the POM to point at the ChemAxon Repo for obtaining the JAR
- Made changes to the DockerFile (Needs more review) to remove the local add of the ChemAxon Jar.

## Related Issue
No Related Issues

## How Has This Been Tested?
I wrote one very simple test to load an SDF file, but we should talk about adding this to a more structured unit test for roo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,24 @@
 ARG 	CHEMISTRY_PACKAGE=jchem
-FROM maven:3-openjdk-8 as builder
+FROM maven:3-openjdk-11 as builder
 
 FROM 	builder as dependencies
 ARG     CHEMISTRY_PACKAGE
 ENV     CHEMISTRY_PACKAGE=${CHEMISTRY_PACKAGE}
 
 FROM 	dependencies as jchem
-ADD 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
-RUN     mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
+# ADD 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
+# RUN     mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
 
 FROM 	${CHEMISTRY_PACKAGE} as compile
+ADD settings.xml /root/.m2/settings.xml
 ADD 	pom.xml /src/pom.xml
 WORKDIR /src
-RUN 	mvn dependency:resolve -P ${CHEMISTRY_PACKAGE}
+RUN 	mvn dependency:resolve -P ${CHEMISTRY_PACKAGE} -X
 ADD 	. /src
 RUN 	mvn clean && \
-        mvn compile war:war -P ${CHEMISTRY_PACKAGE}
-
-FROM tomcat:9.0.62-jre8-openjdk-slim-buster
+        mvn compile war:war -P ${CHEMISTRY_PACKAGE} -X
+# RUN mvn compile war:war -P ${CHEMISTRY_PACKAGE}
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -58,6 +58,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.]]>
 		<url>https://github.com/mcneilco/acas-roo-server.git</url>
 	</scm>
 	<repositories>
+		<repository>
```
bffrost, line 61: Could this `<repository>` be moved into the `jchem` build profile so it isn't contacted during non-jchem builds?

pom.xml
```diff
@@ -33,7 +33,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.]]>
 	<description>ACAS Java persistence layer</description>
 	<properties>
 		<aspectj.version>1.9.7</aspectj.version>
-		<java.version>1.8</java.version>
+		<java.version>11</java.version>
```
bffrost, line 36: This is a big change. Was this required for the new JChem version?

bffrost:
@philippcheung for testing, it'd be useful to run through the integration tests in [acasclient](https://github.com/mcneilco/acasclient). Could you get a local instance of ACAS `release/2022.1.x` running with this new version of JChem and confirm that the acasclient tests run successfully?

Dockerfile
```diff
@@ -1,23 +1,24 @@
...
+FROM maven:3-openjdk-11 as builder
 
 FROM 	builder as dependencies
 ARG     CHEMISTRY_PACKAGE
 ENV     CHEMISTRY_PACKAGE=${CHEMISTRY_PACKAGE}
 
 FROM 	dependencies as jchem
-ADD 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
-RUN     mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
+# ADD 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
```
bffrost, line 9: Previously we'd been assuming this jar file would be supplied despite it not being in the repo. Perhaps you could do something similar for the API keys?

Dockerfile
```diff
@@ -1,23 +1,24 @@
...
+# RUN     mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
 
 FROM 	${CHEMISTRY_PACKAGE} as compile
+ADD settings.xml /root/.m2/settings.xml
```
bffrost, line 13: What's this file? I don't see it in the repo. Is this the API key?

Dockerfile
```diff
@@ -1,23 +1,24 @@
...
 ADD 	pom.xml /src/pom.xml
 WORKDIR /src
-RUN 	mvn dependency:resolve -P ${CHEMISTRY_PACKAGE}
+RUN 	mvn dependency:resolve -P ${CHEMISTRY_PACKAGE} -X
```
bffrost, line 16: I don't think we need the maven debug flag on permanently. Can this be removed once you've finished debugging?

bffrost:
@philippcheung we're planning on putting out a release of ACAS for Schrodinger in the next week or so. We're currently undergoing QA and putting in a few targeted bug fixes. Please hold off on merging this major change until we have a chance to put out the `2022.1.0`.


It looks like this is missing a database migration too for the JChem upgrade: https://github.com/mcneilco/acas-roo-server/tree/release/2022.1.x/src/main/java/com/labsynch/labseer/db/migration/postgres/jchem 
If you want to break up the JChem upgrade work into multiple PRs, then perhaps you could make an intermediate branch like `jchem-22.12.0` and then merge a series of PRs into that branch until it's stable, rather than merging each into `release/2022.1.x` and potentially breaking the release branch.

PR #350, opened 2022-08-12
Update rename batch methods to copy concentration and conc_units

brianbolt:
## Description
Rename batch (part of reparent lot) was not properly copying concentration and concentration unit fields when creating the new batch code value after ignoring the old one.  This fixes that.

## Related Issue
ACAS-402

## How Has This Been Tested?
Locally, ran reparent lot on a lot with dose response data and verified that concentration data was now copied and that curves rendered properly.
Verified that this endpoint which return raw data to the curve rendering service properly ignores `batch code` values which have been ignored in the subject value table.

40 rows in dose response file, 40 rows returned from services (80 rows without this fix).
```
curl --location --request POST 'http://localhost:8080/acas/api/v1/curvefit/rawdata?format=tsv&respoonse=efficacy' \
--header 'Content-Type: application/json' \
--header 'Cookie: connect.sid=s%3AKNQqyEFTfXahf2nBQUOA0yUpr8Ih7VP5.rEi8p2ZVFTdLBA%2Bk8EhE%2BwCzYg%2FGKSY0EJmjE%2FWnT%2F0' \
--data-raw '[
    "AG-00000070_23"
]'
```


bffrost (approved):
LGTM

src/main/java/com/labsynch/labseer/dto/RawCurveDataDTO.java
```diff
@@ -340,6 +341,7 @@ public static List<RawCurveDataDTO> getRawCurveDataAgonist(Collection<String> cu
 				+ "AND rsv.lsKind = :responseKind "
 				+ "AND bcsv.lsType = 'codeValue' "
 				+ "AND bcsv.lsKind = 'agonist batch code' "
+				+ "AND bcsv.ignored = false "
```
brianbolt, line 344: This is not used right now but I decided to fix it when I saw it here as well.

bffrost (approved):
LGTM

PR #351, opened 2022-08-15
ACAS-403: Fix lack of molWeight updates when swapping structures

bffrost:
## Description
Parent structure swap route was not updating molecular weight, mol formula, lot molweights, or structure tables.
Switched to using existing parentStructureService.update method which does all this.

## How Has This Been Tested?
Added additional acasclient tests to confirm mol weight and mol formula changes. Confirmed those tests were broken before this change and fixed afterwards.

Structure table updates will need to be tested on a BBChem system since indigo does not use a separate structure table AFAIK.


brianbolt (approved):
LGTM

PR #353, opened 2022-08-29
ACAS 373 - Implement Maven Dependency Caching 

dalejerikson:
**Description**

acas-roo-server Docker builds on GitHub Actions currently take 10 minutes on average.

A portion of this is re-downloading java package dependencies on every build.

The goal of this branch is to cache dependencies to improve build performance.


.github/workflows/docker-image.yml
```diff
@@ -16,6 +16,24 @@ jobs:
...
+          java-version: 11
+      # Cache Save and Restore
+      - name: Java Dependency Cache Step
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      # Run Maven Tests
```
bffrost, line 34: These tests are not well maintained. I don't think we should be running them as part of our CI at this point.
If you remove this step, does the build pass?

bffrost (changes requested):

.github/workflows/docker-image.yml
dalejerikson: I didn't have them prior and it worked. When I was doing some knowledge digging, I saw people adding `mvn test` as a best practice. If it's unstable, I understand keeping it out. I just reran the pipeline without the tests though and there seems to be a generic issue. The steps for setting up JDK and and caching do pass. This error is at the `Set up QEMU` step. 

```Pulling binfmt Docker image
  /usr/bin/docker pull tonistiigi/binfmt:latest
  Error response from daemon: Get "https://registry-1.docker.io/v2/": EOF
  Error: The process '/usr/bin/docker' failed with exit code 1

bffrost: Yeah, makes sense. It is a best practice to run unit tests, provided we have a working unit test suite. I expect those tests to fail, but we could always investigate and see. I think the error you posted (JChem dependency not found) is expected if we run `mvn test` with no other arguments, because it'll try to use the default JChem build profile. But you might be able to get further by passing the `-P ${CHEMISTRY_PACKAGE}` argument to specify either the bbchem or indigo flavors.
See the Dockerfile which runs `mvn dependency:resolve -P ${CHEMISTRY_PACKAGE}`


That generic issue looked like a transient issue with access to some public resource, so I reran your jobs and they seem to have passed this time: https://github.com/mcneilco/acas-roo-server/runs/8076853823?check_suite_focus=true

bffrost: @dalejerikson I was curious so I re-ran those jobs another time, to see if they would be faster. Unfortunately they didn't seem to go much faster, and the logging indicates that it was re-downloading maven dependencies.
I tried to verify using the GitHub Actions Cache API (https://docs.github.com/en/rest/actions/cache#list-github-actions-caches-for-a-repository)
But https://api.github.com/repos/mcneilco/acas-roo-server/actions/cache/usage indicates there are no caches for this repo.
Could you dig in further and try to get a GitHub actions run where it is clearly using the cache and going faster?

dalejerikson: Screen Shot 2022-08-30 at 9 55 06 AM
Hmmmm wonder if this, at least in part, is the cause.

bffrost: Interesting. You could try adding a pull request trigger to this workflow, at least temporarily. That should make it so that you can push commits to this PR and it'll re-trigger builds.

dalejerikson:
Added Docker Layer Caching and Build Time was 33s/38s respectively: https://github.com/mcneilco/acas-roo-server/runs/8256941229?check_suite_focus=true

bffrost:
@dalejerikson I'm curious looking at the GitHub Actions history.
Your commit "Cleanup" ran super fast: https://github.com/mcneilco/acas-roo-server/actions/runs/3016289315
But then you next commit "Consistency" took the regular 15 minutes: https://github.com/mcneilco/acas-roo-server/actions/runs/3017591790

Is that expected? I'm a little surprised by that.

dalejerikson:
> @dalejerikson I'm curious looking at the GitHub Actions history. Your commit "Cleanup" ran super fast: https://github.com/mcneilco/acas-roo-server/actions/runs/3016289315 But then you next commit "Consistency" took the regular 15 minutes: https://github.com/mcneilco/acas-roo-server/actions/runs/3017591790
> 
> Is that expected? I'm a little surprised by that.

Interesting. I had a few runs that were working and quite fast. The run you mentioned being slower still took a while even when triggering a rebuild (which is definitely a case where it shouldn't take that long). I'll cast a wide net and setup a few more tests to see if I can hone in on what the problem is and if it's a one off or not.

brianbolt:
Since this has been backlogged. I'm going to move this to draft to make it more obvious that its not ready for review.

bffrost:
I'm going to close this PR since this PR is quite stale and we aren't planning to work on this anytime soon.

PR #356, opened 2022-08-30
Update ApiSaltController.java

dalejerikson:
Hotfix to allow `dryrun` parameter to be added with recent Salt Browser development but still allow backwards compatibility and not break previously written unit tests

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiSaltController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiSaltController.java
@@ -305,7 +305,6 @@ public class ApiSaltController {
 
 			return new ResponseEntity<String>(ErrorMessage.toJsonArray(errors), headers, HttpStatus.BAD_REQUEST);
 		} else if (salt.getCdId() > 0 || (validSalt & dryrun)) { // Continue If Valid Registered or Valid Salt in Dryrun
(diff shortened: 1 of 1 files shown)


brianbolt (changes requested):
I think you also need to move line 308 down to just above line 321 and delete line 319 so that you return json with the id filled in when dry run is false.

dalejerikson:
> I think you also need to move line 308 down to just above line 321 and delete line 319 so that you return json with the id filled in when dry run is false.

Thanks for the catch. Just updated.

brianbolt (approved):
LGTM if tests are passing

dalejerikson:
> LGTM if tests are passing

There's a failure in `test_045_register_sdf_case_insensitive` where `self.assertEqual(len(warnings), 6)` fails. The number of warnings seems to be 0. These warnings are obtained from `response = self.client.register_sdf(upload_file_file, "bob", mappings, dry_run=True)`. Not sure what these warnings are supposed to be. There's no errors though related to calling the salt routes.

bffrost (approved):
LGTM

bffrost:
@dalejerikson I think you're good to merge here. We can make follow-up fixes to that remaining test, but I agree with you that it doesn't look related to this.

PR #358, opened 2022-09-06
ACAS-6: Switch to generic V3000/V2000 mol reader

bffrost:
## Description
Fixing an issue mentioned in https://github.com/mcneilco/acas/pull/1002
The V2000 reader cannot read V3000 MOL strings, whereas this implementation works for both V3000 and V2000 MOL strings.

## How Has This Been Tested?
Tested in local dev setup and confirmed both V3000 and V2000 inputs generated images successfully.


dalejerikson (approved):
👍

PR #359, opened 2022-09-06
ACAS-6: Follow-up fix for image rendering

bffrost:
## Description
Add a new route to render a MOL string as base64-encoded PNG.
Also fixed a small grammar issue in duplicate messages.

## How Has This Been Tested?
Tested locally alongside changes from https://github.com/mcneilco/acas/pull/1002

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiStructureImageController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiStructureImageController.java
@@ -1,5 +1,6 @@
 package com.labsynch.labseer.api;
 
+import java.util.Base64;
 import java.util.List;
 
 import com.labsynch.labseer.domain.Lot;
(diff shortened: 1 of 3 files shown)


src/main/java/com/labsynch/labseer/api/ApiStructureImageController.java
```diff
@@ -60,6 +61,29 @@ public ResponseEntity<byte[]> convertMol(@RequestBody String molStructure,
 
     }
 
+    @RequestMapping(value = "/convertMol/base64", method = RequestMethod.POST)
```
bffrost, line 64: This is a simple derivative of the existing `convertMol` route immediately above. All I added was the two lines:
```
Base64.Encoder encoder = Base64.getEncoder();
String encodedString = encoder.encodeToString(image);
```
and changed the return type to `String` instead of `bytes[]`

dalejerikson (approved):
👍

PR #360, opened 2022-09-09
DE Docker Optimization 

dalejerikson:
Reviewed the following checklist, validated presence of best practice, and made minor modifications. The build time and size of the image are reasonable. The longest steps are downloading dependencies/packages so caching these (whether locally, on GitHub, etc.) will significantly decrease time to build. The two changes I made to align with best practices are: 

- Replace ADD with COPY where applicable in order to prevent path traversal and zip slip vulnerabilities 
- Organize and sort package dependencies alphanumerically 

I experimented with moving RUN and ADD steps around. I wanted to move long running, infrequently changed steps to higher up in the file(s). However, this yielded insignificant improvements. I also tried minimizing the number of layers but that would increase the likelihood of a cache being invalidated. In looking at build times, most of the steps are trivial in the amount of time that they took and the file is mostly organized in the required sequential order steps need to be conducted in. 

### Background & Context
Docker Images Layers
- [x] Images are composed of a series of layers.
- [x] The RUN, ADD, & COPY build steps create new layers.
**Docker Image Cache**
- [x] Docker caches the results of each build step.
- [x] On cache hits, there's no need to run an instruction. This is useful for skipping instructions like apt-get install which may take some 
- [x] Move long running steps which change infrequently higher up in your Dockerfile.
- [x] For example, installing dependencies w/ apt-get, pip, or yarn
**Downloading Images**
- [x] Image layers are uploaded to a central repository and then downloaded to k8s nodes as needed.
- [x] Smaller images download faster, meaning pods start faster.
- [x] Smaller images also result in less network traffic and disk usage.
- [x] Image layers can be cached individually on a k8s node.
### Build Context
- [x] The build context is the set of files available for copying & adding to the container image.
- [x] The context defaults to Dockerfile's folder (& subtree) All file in the context are sent to Docker daemon. 
- [x]  Since Docker is a daemon and may be called across a network, it is designed so files are sent to it rather than accessed as needed off disk. Thus a larger context will result in a slower build.
- [x] A .dockerignore file exists to configure excluding files from the build context. The .dockerignore file works like .gitignore and many .gitignore best practices apply.
- [x] Minimizing the context can also reduce the risk of accidentally including unintended files such as credentials.
### Multistage Builds
- [x] Multistage builds allow copying specific files from early stages to later stages.  The final container won't include any image layers from earlier stages. This means smaller images with less layers.
- [x] Multistage builds can be used to remove build dependencies.
- [x] If you're building software within a Dockerfile, you can build in one stage and publish the artifacts in a 2nd stage.
- [x] In that scenario, all the dependencies for the build will discarded, reducing the final image size.
- [x] This approach also minimizes the risk of credentials leaking if needed for the build step.
### Tips & Best Practices
Performance
- [x] Order matters for caching.  Changes to earlier commands will cause later commands to be invalidated and re-run. 
- [x] Move long running steps which change infrequently higher up in your Dockerfile.
- [x] For example, installing dependencies w/ apt-get, pip, or yarn
- [x] Changes to copied files will break the cache.  Copy the minimum number of files needed for before a Dockerfile step.
- [x] Think about a RUN command as a cacheable unit. 
- [x] A large command will be invalidated more frequently. 
- [x] Smaller commands create overhead and may use unnecessary disk space for intermediate layers.
- [x] Use .dockerignore to exclude files from the build context. 
- [x] This can speed up builds immensely since large directory trees like node_modules can be ignored and won't need to be scanned when creating the build context.
- [x] Minimizing the context can also reduce the risk of accidentally including unintended files such as credentials.
### Image Size
- [x] Split out multistage builds to remove build dependencies.  If you're building software within a Dockerfile, build in one stage and publish the artifacts in a 2nd stage.  This can also reduce the risk of accidentally including credentials needed during the build stage.
- [x] Minimize the number of layers to reduce overhead.  Combine RUN, ADD, or COPY commands This can be a trade-off with thinking about RUN as a cacheable unit and leveraging the build cache.
- [x] Check your docker image size by using docker history <docker-image-hash> 
**Security**
- [x] Run commands in containers as a user with the minimal necessary privileges.
- [x] Don't put sensitive information into a Docker image.  Use features designed for sharing secrets instead.
- [x] Use COPY instead of ADD whenever possible.  This prevents path traversal & zip slip vulnerabilities.
- [x] Restrict resources (RAM, CPU) & capabilities
### Maintainability
- [x] Use official images where possible. 
- [x] Use specific image tags rather than latest. 
- [x] Run only a single process in each container. 
### Data
- [x] Store persistent data on volumes. 
- [x] Data is persisted and it's easier to provision disk space
- [x] Data can be backed up
- [x] Avoid storing application data in image layers.
- [x] Writes in running containers use disk on host machine
- [x] Data is lost on container restart.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,22 +14,39 @@ ADD 	pom.xml /src/pom.xml
 WORKDIR /src
 RUN 	mvn dependency:resolve -P ${CHEMISTRY_PACKAGE}
 ADD 	. /src
+
+# Slow Step - Caching this Layer Can Greatly Improve Build Time  (~6x Faster)
 RUN 	mvn clean && \
         mvn compile war:war -P ${CHEMISTRY_PACKAGE}
 
 FROM tomcat:9.0.62-jre8-openjdk-slim-buster
 
+# Third and Last Step That Requires Significant (Relative) Amount of Time 
 RUN apt-get update && \
-    apt-get install -y openssl libfontconfig libfreetype6 curl
+    apt-get install -y \
+      curl \
+      libfontconfig \
+      libfreetype6 \
+      openssl 
 
 RUN  sed -i 's/<Connector port="8080"/<Connector address="${listen.address}" port="8080"/' conf/server.xml
 
 # Add nodejs for prepare config files
 ENV NPM_CONFIG_LOGLEVEL warn
 ENV NODE_VERSION 14.x
+
+# Second Slowest Step 
 RUN curl -fsSL https://deb.nodesource.com/setup_$NODE_VERSION | bash - && \
   apt-get install -y nodejs && \
-  npm install -g coffeescript@2.5.1 properties@1.2.1 underscore@1.12.0 underscore-deep-extend@1.1.5 properties-parser@0.3.1 flat@5.0.2 glob@7.1.6
+  npm install -g \
+    coffeescript@2.5.1 \
+    flat@5.0.2 \ 
+    glob@7.1.6 \ 
+    properties@1.2.1 \
+    properties-parser@0.3.1 \
+    underscore@1.12.0 \ 
+    underscore-deep-extend@1.1.5 
+
 ENV NODE_PATH /usr/lib/node_modules
 
 # Add runner user so we don't run as root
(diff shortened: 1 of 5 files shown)


bffrost (approved):
LGTM! Thanks for the analysis @dalejerikson . Looks like we're already pretty clean, and the work you're already doing on caching will help speed up our builds.

PR #361, opened 2022-09-13
Hotfix - Editing Salt Treats No Change to Abbrev / Name As Duplicate in System 

dalejerikson:
**Reproduce/Test:**

As a CReg admin, navigate to the CmpdRed Salts panel, pick any salt from the table and edit the abbreviation or name.
Click the 'Next' button.

**Bug Results:**

Any changes cannot be saved since they are always referred to as duplicates of existing values, even though they are added as a part of the same workflow.

**Post-Fix Results:**

After reviewing the dependencies, the changes should be allowed to save. There should be no error/warning for the name OR abbreviation already existing in the system if the name OR abbreviation (respectively) of the active salt edit isn't changed.

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
@@ -245,26 +245,35 @@ public class SaltServiceImpl implements SaltService {
 			warnings.add(warning);
 		}
 
-		List<Salt> saltsByName = Salt.findSaltsByNameEquals(newSalt.getName()).getResultList();
-		if (saltsByName.size() > 0 && !newSalt.getName().equals(oldSalt.getName())) {
-			ErrorMessage error = new ErrorMessage();
-			error.setLevel("error");
-			error.setMessage("Duplicate salt name. Another salt exists with the same name.");
-			warnings.add(error);
+		if(!newSalt.getName().equals(oldSalt.getName()))
+		{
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
```diff
@@ -246,21 +246,21 @@ public ArrayList<ErrorMessage> validateSaltEdit(Salt oldSalt, Salt newSalt)
 		}
 
 		List<Salt> saltsByName = Salt.findSaltsByNameEquals(newSalt.getName()).getResultList();
-		if (saltsByName.size() > 0) {
+		if (saltsByName.size() > 0 && !newSalt.getName().equals(oldSalt.getName())) {
```
bffrost, line 249: nit: I think this will work, but the logic is a little confusing to me. We're doing a search, then ignoring the results if the salt name didn't change.
Could we skip the search entirely if the salt name hasn't changed?

Or the other approach that makes sense to me is to always do the search, then filter out the "current salt" from the results by ID, and see what (if any) other results are left.

bffrost (approved):

src/main/java/com/labsynch/labseer/service/SaltServiceImpl.java
dalejerikson: I like the idea of skipping the search if there hasn't been a change. If the salt was allowed to be registered in the first place with the name and/or abbreviation then there shouldn't be another so the search would theoretically be ineffective.

bffrost (approved):
LGTM

PR #362, opened 2022-09-14
ACAS-400: Fix error flow for duplicate parent aliases

bffrost:
## Description
- This is meant to fix a bug when CReg is configured to reject duplicate aliases and you try to register a new compound with a duplicate alias. Currently CReg will save the parent, then return a 500 server error. Locally the error gets rendered properly in the client, but gives the false impression that the parent wasn't registered. In typical LiveDesign deployment scenarios, the 500 error is trapped by nginx and the CReg client can't properly handle & display the cause of the error.
- Added a call to "validateParentAliases" within the MetaLot processAndSave route _before_ the parent is saved. Previously it only happened as part of "updateParentAliases" which happened after the parent was saved.
- Switched the error response of MetaLot from 500 to 400. Along the way I discovered that the ACAS node proxy actually checks the content of the response for the word "Duplicate" and amends the status code to 409 if it finds that, but otherwise passes the status through. For all of the caught errors in MetaLot saving, a 400 "bad request" error which indicates bad inputs is more accurate than 500 "internal server error" which indicates an unexpected backend failure.

## How Has This Been Tested?
Tested in local development. Will need to verify this in LD deployments.

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiMetalotController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiMetalotController.java
@@ -117,7 +117,7 @@ public class ApiMetalotController {
 		if (hasError) {
 			headers.add("lot save error", "true");
 			return new ResponseEntity<String>(ErrorMessage.toJsonArray(mr.getErrors()), headers,
-					HttpStatus.INTERNAL_SERVER_ERROR);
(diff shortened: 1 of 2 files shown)


dalejerikson (approved):
👍

iamanandkr (approved):
Great description!

PR #363, opened 2022-09-15
ACAS-439: Performance rewrite of checkPurgeFileDependencies

bffrost:
## Description
We'd never optimized the performance of the checkPurgeFileDependencies route, and it was written very naively, with a lot of unnecessary and repeated fetching of data. Rewrote it to do a single DB query to pull back only the minimal relevant parent / saltForm / lot information as a flattened set of DTO objects, then loop through and build up the same data structures as before.

## How Has This Been Tested?
Tested locally both for performance and for correct behavior.
Performance for checking dependencies on a 999 lot file with minimal dependencies went from 2.5 minutes to < 1 s.
Confirmed that single registered lots are correctly caught and reported.
Ran ACASClient test that checks SEL file dependencies, and that test passed.

### DIFF
--- a/src/main/java/com/labsynch/labseer/dto/BulkLoadParentSaltFormLotDTO.java
+++ b/src/main/java/com/labsynch/labseer/dto/BulkLoadParentSaltFormLotDTO.java
@@ -1,0 +1,108 @@
+package com.labsynch.labseer.dto;
+
+import java.util.Collection;
+import java.util.List;
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+import flexjson.JSONDeserializer;
+import flexjson.JSONSerializer;
+
+public class BulkLoadParentSaltFormLotDTO {
+
+    private String parentCorpName;
+
+    private Long parentBulkLoadFileId;
+
+    private String saltFormCorpName;
+
+    private Long saltFormBulkLoadFileId;
+
+    private String lotCorpName;
+
+    private Long lotBulkLoadFileId;
+
+    public BulkLoadParentSaltFormLotDTO(String parentCorpName, Long parentBulkLoadFileId, String saltFormCorpName, Long saltFormBulkLoadFileId, String lotCorpName, Long lotBulkLoadFileId) {
+        this.parentCorpName = parentCorpName;
+        this.parentBulkLoadFileId = parentBulkLoadFileId;
+        this.saltFormCorpName = saltFormCorpName;
+        this.saltFormBulkLoadFileId = saltFormBulkLoadFileId;
+        this.lotCorpName = lotCorpName;
+        this.lotBulkLoadFileId = lotBulkLoadFileId;
+    }
+
+    public BulkLoadParentSaltFormLotDTO() {
+
+    }
+
+    public String toJson() {
+        return new JSONSerializer()
+                .exclude("*.class").serialize(this);
+    }
+
+    public static String toJsonArray(Collection<BulkLoadParentSaltFormLotDTO> collection) {
+        return new JSONSerializer()
+                .exclude("*.class").serialize(collection);
+    }
+
+    public static BulkLoadParentSaltFormLotDTO fromJsonToBulkLoadParentSaltFormLotDTO(String json) {
+        return new JSONDeserializer<BulkLoadParentSaltFormLotDTO>()
+                .use(null, BulkLoadParentSaltFormLotDTO.class).deserialize(json);
+    }
+
+    public static Collection<BulkLoadParentSaltFormLotDTO> fromJsonArrayToBulkLoadParentSaltFormLotDTOs(String json) {
+        return new JSONDeserializer<List<BulkLoadParentSaltFormLotDTO>>()
+                .use("values", BulkLoadParentSaltFormLotDTO.class).deserialize(json);
+    }
+
+    public String getParentCorpName() {
+        return this.parentCorpName;
+    }
+
+    public void setParentCorpName(String parentCorpName) {
+        this.parentCorpName = parentCorpName;
+    }
+
+    public Long getParentBulkLoadFileId() {
+        return this.parentBulkLoadFileId;
+    }
+
+    public void setParentBulkLoadFileId(Long parentBulkLoadFileId) {
+        this.parentBulkLoadFileId = parentBulkLoadFileId;
+    }
+
+    public String getSaltFormCorpName() {
+        return this.saltFormCorpName;
+    }
+
+    public void setSaltFormCorpName(String saltFormCorpName) {
+        this.saltFormCorpName = saltFormCorpName;
+    }
+
+    public Long getSaltFormBulkLoadFileId() {
+        return this.saltFormBulkLoadFileId;
+    }
+
+    public void setSaltFormBulkLoadFileId(Long saltFormBulkLoadFileId) {
+        this.saltFormBulkLoadFileId = saltFormBulkLoadFileId;
+    }
+
+    public String getLotCorpName() {
+        return this.lotCorpName;
+    }
+
+    public void setLotCorpName(String lotCorpName) {
+        this.lotCorpName = lotCorpName;
+    }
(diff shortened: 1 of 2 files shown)


src/main/java/com/labsynch/labseer/dto/BulkLoadParentSaltFormLotDTO.java
```diff
@@ -0,0 +1,108 @@
+package com.labsynch.labseer.dto;
+
+import java.util.Collection;
+import java.util.List;
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+import flexjson.JSONDeserializer;
+import flexjson.JSONSerializer;
+
+public class BulkLoadParentSaltFormLotDTO {
```
bffrost, line 10: This "Data Transfer Object" (DTO) class is just a data container that we'll use to collect the results of a SQL (really JPQL) query. These are also known as "POJO"s (Plain Old Java Objects).

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2051,84 +2055,107 @@ public String generateSummaryText(int numRecordsRead,
 		return summary;
 	}
 
+	public Collection<BulkLoadParentSaltFormLotDTO> getBulkLoadDTOsByBulkLoadFileId(Long bulkLoadFileId) {
+		// This query pulls back the corp name and bulk load file ids for all parents, saltforms, and lots that are at all linked
+		// to the bulkLoadFileId in question. 
+		EntityManager em = Parent.entityManager();
+        TypedQuery<BulkLoadParentSaltFormLotDTO> q = em.createQuery("SELECT new com.labsynch.labseer.dto.BulkLoadParentSaltFormLotDTO("
```
bffrost, line 2062: Here's the main query that allows us to skip a lot of repeat work. We're grabbing just the corpName and bulkLoadFile IDs from the parent, saltForm, and lot tables, where any of the three (parent, saltform, or lot) are tied to the current bulkLoadFile that's being assessed for purging.

This means we'll get some repeated values in some columns, for example if a parent has three lots, one from the same file, one from a different file, and one from single registration (bulkLoadFile = null).

In SQL this query equates to:
```
select p.corp_name, p.bulk_load_file, s.corp_name, s.bulk_load_file, l.corp_name, l.bulk_load_file
from parent p
join salt_form s on p.id = s.parent
join lot l on l.salt_form = s.id
where p.bulk_load_file = 2091 or s.bulk_load_file = 2091 or l.bulk_load_file = 2091;
```
where 2091 is an example bulkLoadFileId.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2051,84 +2055,107 @@ public String generateSummaryText(int numRecordsRead,
...
+		// cmpdRegDependencies is a map of corpName : list of CReg BulkLoadFiles that depend on that corpName
 		Map<String, HashSet<String>> cmpdRegDependencies = new HashMap<String, HashSet<String>>();
+		// dependentSingleRegLots is a unique list of lot corpnames which came from single (non-bulk) registration and depend on parents or saltforms from this file
 		HashSet<String> dependentSingleRegLots = new HashSet<String>();
-		// we make a map with a CorpName as each key, and a list of dependent CorpNames
-		// as each value
-		int numberOfParents = 0;
-		int numberOfSaltForms = 0;
-		int numberOfLots = 0;
-		Collection<Parent> parents = Parent.findParentsByBulkLoadFileEquals(bulkLoadFile).getResultList();
```
bffrost, line 2064: There are a lot of changes here, sorry! The main gist of this is:
Before: pull a list of Parents linked to this bulkLoadFile, then loop through its children and children's children, adding to the "acasDependencies", "cmpdRegDependencies", and "dependentSingleRegLots" Maps as we go.
Then repeat the same process with SaltForms (and their children), and then finally with Lots.

After: Do one big fetch of data using `getBulkLoadDTOsByBulkLoadFileId`, giving us all the info we need, then do some careful checking of the bulkLoadFile id columns for parent, saltform, and lot, and build up the same Map datastructures.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2051,84 +2055,107 @@ public String generateSummaryText(int numRecordsRead,
...
+		Map<Long, String> bulkLoadIdsToNames = new HashMap<Long, String>();
+		
+		for (BulkLoadParentSaltFormLotDTO dto : dtos) {
+			String parentCorpName = dto.getParentCorpName();
+			String saltFormCorpName = dto.getSaltFormCorpName();
+			String lotCorpName = dto.getLotCorpName();
+			Long parentBulkLoadFileId = dto.getParentBulkLoadFileId();
+			Long saltFormBulkLoadFileId = dto.getSaltFormBulkLoadFileId();
+			Long lotBulkLoadFileId = dto.getLotBulkLoadFileId();
+			// acasDependencies, unique corpName lists: collect all parent, saltform, and lot corpnames directly tied to the file
```
bffrost, line 2110: I've tried to more clearly document what each of these Maps are supposed to contain, so the logic below is easier to follow.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2051,84 +2055,107 @@ public String generateSummaryText(int numRecordsRead,
...
+        TypedQuery<BulkLoadParentSaltFormLotDTO> q = em.createQuery("SELECT new com.labsynch.labseer.dto.BulkLoadParentSaltFormLotDTO("
+		+ "p.corpName, p.bulkLoadFile.id, sf.corpName, sf.bulkLoadFile.id, l.corpName, l.bulkLoadFile.id) "
+		+ "FROM Lot l JOIN l.saltForm sf JOIN sf.parent p "
+		+ "WHERE l.bulkLoadFile.id = :bulkLoadFileId OR sf.bulkLoadFile.id = :bulkLoadFileId OR p.bulkLoadFile.id = :bulkLoadFileId",
+		BulkLoadParentSaltFormLotDTO.class);
+        q.setParameter("bulkLoadFileId", bulkLoadFileId);
+        return q.getResultList();
+	}
+
+	public Map<String, HashSet<String>> addToHashMapAsList(String key, String value, Map<String, HashSet<String>> map){
```
bffrost, line 2071: This was just a utility function to avoid repeat logic. In Python we'd use a defaultdict, but I'm not sure what the Java equivalent is, so I just put this logic into a helper function.

dalejerikson (approved):
Nice, clean, easy to follow!

iamanandkr (approved):
Great description and comments!

brianbolt (approved):
A few small indent nits but overall LGTM.  Thanks for this refactor.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2051,84 +2055,107 @@ public String generateSummaryText(int numRecordsRead,
...
+			HashSet<String> values = new HashSet<String>();
+			values.add(value);
+			map.put(key, values);
+		}
+		return map;
+	}
+
 	@Override
 	public PurgeFileDependencyCheckResponseDTO checkPurgeFileDependencies(BulkLoadFile bulkLoadFile) {
+		// acasDependencies is a map of corpName : list of ACAS experiments that depend on that corpName
```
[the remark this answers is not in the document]
bffrost: Thanks. I was using list carelessly. They are unique lists, really sets. I'll update the comments.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
bffrost: Thanks, fixed it.

PR #366, opened 2022-09-26
Update swapParentStructures to support parent aliases.

iamanandkr:
Update `swapParentStructures` API to support parent aliases. If the API user passed a parent alias name instead of a corporate name to the `swapParentStructures` API it fails with the error,

```
"Caught error while fetching parent"
```

Goal here is to support unique parent aliases i.e if no parent is found matching the corporate name then fallback to finding if there is a unique parent alias matching the name. If a unique parent alias exists then use the parent as the object to swap structures.

## Description

Create a private API that handles the logic for returning the parent based on the criteria.

## Related Issue

[ACAS-440](https://jira.schrodinger.com/browse/ACAS-440)

## How Has This Been Tested?

Added acasclient tests to check the behaviour.

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
@@ -2,7 +2,7 @@ package com.labsynch.labseer.service;
 
 import java.util.List;
 import java.util.ArrayList;
-import java.util.Date;
+import java.util.Collections;
 import java.util.HashSet;
 
 import org.slf4j.Logger;
@@ -44,7 +44,7 @@ public class ParentSwapStructuresServiceImpl implements ParentSwapStructuresServ
 		try {
 			return Parent.findParentsByCorpNameEquals(name).getSingleResult();
 		} catch (Exception e) {
-			logger.warn("No parent found for {}", name);
+			logger.warn(String.format("No parent found for corporate id - '%s'", name), e);
 		}
 
 		// Check if a unique alias with the name exists.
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
```diff
@@ -29,6 +32,36 @@ public class ParentSwapStructuresServiceImpl implements ParentSwapStructuresServ
...
+			parentAliases = ParentAlias.findParentAliasesByAliasNameEquals(name).getResultList();
+		} catch (Exception e) {
+			logger.warn("No parent aliases found for {}", name);
+		}
+		if (parentAliases.size() == 1) {
+			ParentAlias parentAlias = parentAliases.get(0);
+			logger.info("Unique parent alias found for {}",  name);
+			return parentAlias.getParent();
+		}
+		throw new ParentNotFoundException("No parent or unique alias found for " + name);
```
bffrost, line 62: I think it'd actually be better to split this out and throw a different message depending on whether `parentAliases.size()` is 0 or > 1. You have the information at your fingertips to give a more complete message, for example providing the corporate IDs if there are multiple duplicate aliases (size>1), or communicating a clearer message that the name was not found as a corporate id OR an alias (size=0)

bffrost (changes requested):

src/main/java/com/labsynch/labseer/service/ParentSwapStructuresServiceImpl.java
iamanandkr: Yep, I agree. I have updated the error and log messages. Let me know if the messages are appropriate now!

bffrost (approved):
LGTM!

PR #369, opened 2022-09-29
Update swapParentStructures to raise BAD_REQUEST for invalid user inputs.

iamanandkr:
## Description

**Issue:**
When `swapParentStructures` API is called via LiveDesign with an invalid corporate name or when there are validation errors or when swap is going to introduce duplicates the response incorrectly mentions that LiveDesign is restarting.

**Bug:**
In `swapParentStructures` API when the swap doesn't go through we set the response status code to `INTERNAL_SERVER_ERROR` which LiveDesign treats as "an internal error".

**Fix:**
We didn't actually have any internal errors on the server but the client gave bad inputs and so we now change the response status to be `BAD_REQUEST`.

## Related Issue

[ACAS-430](https://jira.schrodinger.com/browse/ACAS-430)

## How Has This Been Tested?

Checked on an actual LiveDesign server that response json reaches the client.
Update acasclient logic and ran all acasclient tests.

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiParentController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiParentController.java
@@ -94,7 +94,7 @@ public class ApiParentController {
 		HashMap<String,Object> response = new HashMap<>();
 		response.put("hasError", hasError);
 		response.put("errorMessage", errorMessage);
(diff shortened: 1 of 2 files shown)


bffrost (approved):
LGTM

PR #370, opened 2022-09-30
ACAS-271: Preserve file naming in SDF bulk loader

bffrost:
## Description
ACAS introduces spaces and numbers in parentheses to keep SDF and zip filenames distinct because it drops them all into a common privateUploads folder. This makes life harder for users who may upload the same file multiple times but get different filenames when they re-download that same file from ACAS, or when they download derived zipfiles.

This acas-roo-server change adds an "originalFileName" column to bulk_load_file to track the original file name, and passes it back to the node layer for rendering.

## How Has This Been Tested?
Local testing of uploading the same file many times. Confirmed that the validation download zip, its contents, the registration download zip, and all the links on the purge files page (including the link / summary after purging) all show the original filename, and all download the correct numbered / distinct file.

### DIFF
--- a/src/main/java/com/labsynch/labseer/domain/BulkLoadFile.java
+++ b/src/main/java/com/labsynch/labseer/domain/BulkLoadFile.java
@@ -32,6 +32,9 @@ public class BulkLoadFile {
     @Size(max = 1000)
     private String fileName;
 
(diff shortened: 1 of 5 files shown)


brianbolt (approved):
LGTM

PR #372, opened 2022-10-08
Need to skip ignroed and deleted aliases when checking for duplicates on the parent during bulk load registration

brianbolt:
## Description
 - Ignored aliases were breaking our duplicate alias detection.  This skips aliases which are ignored or deleted from that validation.

## Related Issue
ACAS-477

## How Has This Been Tested?
Registered a compound with an alias of BLAH and then deleted that alias via the GUI which sets it to ignored in the database
Bulk loaded another lot of the same compound and verified that I got the same erroneous error `Within File, Parent Alias BLAH is not unique`
Applied my fix and the error went away


dalejerikson (approved):
👍

hhan-schrodinger (approved):
Looks good to me!

PR #373, opened 2022-10-08
ACAS-476-fill-missing-structure-memory

brianbolt:
## Description
 - Refactor fill missing structures by adding fillMissingParentStructuresByParentIDList and switching transactional block to be on the new function

## Related Issue
ACAS-476

## How Has This Been Tested?
Ran fill missing structures on a server with 1.7 million compounds and the memory never went over 3 Gb


dalejerikson (approved):
Looks good

iamanandkr (approved):
LGTM!

Is it worth the effort to refactor the logic in the two methods to avoid the duplication of lines and instead pass parameters to a common method to substitute depending on the use case?

bffrost (approved):
LGTM, this just shifts the transaction down to the group level instead of one big transaction.
I don't think it's worth doing the refactor that Anand suggests, because the implementations do differ in using the SaltForm vs Parent java classes, and this code is unlikely to be reused elsewhere.

brianbolt:
Changing the base branch to `ACAS-480-epic` as this is one in a series of tickets to fix standardization for large scale CmpdReg systems.

PR #375, opened 2022-10-12
ACAS-481: Get as drawn structure in batch to speed up standardiation

brianbolt:
## Description
 - Replaced function `getOriginallyDrawnAsStructure(Parent parent)` with`getOriginallyDrawnAsStructuresByParentIds(List<Long> parentIds)` to speed up standardization
 - The logic is the same.  Select min lot number with an as drawn structure and use that as the as drawn structure of the parent.

## Related Issue
ACAS-483

## How Has This Been Tested?
Ran a dry run standardization and it sped up the run from `802.86 average speed (rows/min)` to `9916.62 average speed (rows/min)`

### DIFF
--- a/src/main/java/com/labsynch/labseer/domain/Lot.java
+++ b/src/main/java/com/labsynch/labseer/domain/Lot.java
@@ -1,8 +1,10 @@
 package com.labsynch.labseer.domain;
 
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -17,7 +19,6 @@ import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
-import javax.persistence.NoResultException;
 import javax.persistence.OneToMany;
 import javax.persistence.PersistenceContext;
 import javax.persistence.Query;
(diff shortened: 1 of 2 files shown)


src/main/java/com/labsynch/labseer/domain/Lot.java
```diff
@@ -1701,4 +1677,36 @@ public static Collection<Lot> fromJsonArrayToLots(String json) {
...
+        // Convert the result list into a hashmap of parent id to mol structure
+        HashMap<Long, String> parentIdsToMolStructures = new HashMap<Long, String>();
+        List<Object[]> results = q.getResultList();
+        for (Object[] result : results) {
+            // Cast BigInteger value to Long
+            Long parentId = ((BigInteger) result[0]).longValue();
+            parentIdsToMolStructures.put(parentId, (String) result[1]);
+        }
+
+        // Verify that each of the original pIdGroup parent ids exist in the result set
```
iamanandkr, line 1703: `nit-pick`: I think even though this method is only called in context of methods that have the `pIdGroup` naming, we could probably not reference it here to avoid confusions for people looking at this method for the first time? :thinking:

iamanandkr (approved):
LGTM!

dalejerikson (approved):
👍

brianbolt (commented):
Updating comment to m

src/main/java/com/labsynch/labseer/domain/Lot.java
brianbolt: Thanks, missed that when I refactored this into it's own method

PR #378, opened 2022-10-18
ACAS-486: Speed up standardization execution

brianbolt:
## Description
 - Currently we re-standardize all molecules again after dry run during standardization execution.  To speed this up, we can just re-use the standardizationDryRunCompound chemical representation instead of recalculating it again.
   -  This involves removing @Transient from some properties and making them full database properties stored on each of the bbchem tables.
 - Simplify parent look ups by adding join directly to the Parent by making it an explicit property of the standardization dry run compound class
    -  This lets us drop the stereo comment, stereo category and old weight properties from standardization dry run compound
 - Simplify the process of looking up which compounds need standardization by adding a SyncStatus column with `READY`, `NO_CHANGE_REQUIRED`, and `COMPLETE` states.

## Related Issue
ACAS-480

## How Has This Been Tested?
Ran restandardization on large 1.6 million compound system and it greatly improved the performance to about 20K compounds per minute.


src/main/java/com/labsynch/labseer/api/ApiStandardizationServicesController.java
```diff
@@ -144,6 +144,7 @@ public ResponseEntity<String> findDryRunDupeStructs() {
 
 	}
 
+	@Transactional
```
brianbolt, line 147: I had to play a lot with transactional vs not when getting all this working.  I found that @Transactional here was not detrimental and made sense to have so I kept it.

src/main/java/com/labsynch/labseer/domain/AbstractBBChemStructure.java
```diff
@@ -87,13 +87,10 @@ public void setId(Long id) {
 
     private String registrationComment;
 
-    @Transient
```
brianbolt, line 90: Making these not transient makes it so hibernate stores the property in the table which we now use when standardizing to make it so we don't have to call cmpdStructureService a second time when doing the actual standardization after dry run.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -96,7 +89,23 @@ public class StandardizationDryRunCompound {
...
+		READY
+	}
+
+	@ManyToOne
+	@NotNull
+	@JoinColumn(name = "parent_id")
+	private Parent parent;
+	
 	public StandardizationDryRunCompound() {
+		this.setSyncStatus( SyncStatus.NO_CHANGE_REQUIRED);
```
brianbolt, line 108: Defaults to not change required and we set it to READY if during dry run if it needs to be standardized (detects a compound change of some sort).

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -116,32 +125,25 @@ public static TypedQuery<Integer> findMaxRunNumber() {
...
+		return StandardizationDryRunCompound.entityManager().createQuery(querySQL, StandardizationDryRunCompound.class).setParameter("syncStatus", SyncStatus.READY);
 	}
 
-	public static TypedQuery<Long> findParentIdsWithStandardizationChanges() {
-		String querySQL = "SELECT o.parentId FROM StandardizationDryRunCompound o WHERE changedStructure = true OR existingDuplicateCount > 0 OR newDuplicateCount > 0 OR displayChange = true";
-		return StandardizationDryRunCompound.entityManager().createQuery(querySQL, Long.class);
+	@Transactional
+	public static int getReadyStandardizationChangesCount() {
+		String querySQL = "SELECT count(o.id) FROM StandardizationDryRunCompound o WHERE o.syncStatus = :syncStatus and (o.changedStructure = true OR o.existingDuplicateCount > 0 OR o.newDuplicateCount > 0 OR o.displayChange = true)";
+		return StandardizationDryRunCompound.entityManager().createQuery(querySQL, Long.class).setParameter("syncStatus", SyncStatus.READY).getSingleResult().intValue();
```
brianbolt, line 136: These queries are just updated to use our new syncStatus column instead of embedding the restandardizes logic in all of the queries.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -645,12 +648,13 @@ public int recalculateLotMolWeights(List<Long> parentIds) {
...
 	@Transactional
-	public int restandardizeParentStructures(List<Long> parentIds)
+	public int restandardizeParentsOfStandardizationDryRunCompounds()
 			throws CmpdRegMolFormatException, StandardizerException, IOException {
 
 		int batchSize = propertiesUtilService.getStandardizationBatchSize();
 
-		int totalCount = parentIds.size();
+		int totalCount = StandardizationDryRunCompound.getReadyStandardizationChangesCount();
+		int remainingCount = totalCount;
```
brianbolt, line 657: Instead of getting a giant list of parentIds (which can be a lot), we will loop until we have no more standardization dry run compounds in the READY state.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -681,63 +683,46 @@ public int restandardizeParentStructures(List<Long> parentIds)
...
 		}
+		// Get the cmpd reg molecules from the dry run standardization
+		HashMap<String, CmpdRegMolecule> cmpdRegMolecules = chemStructureService.getCmpdRegMolecules(cdIdMap, StructureType.STANDARDIZATION_DRY_RUN);
 
-		// Do standardization
-		logger.info("Starting standardization of " + parentIdsToStructures.size() + " compounds");
-		// Start timer
-		long standardizationStart = new Date().getTime();
-		HashMap<String, CmpdRegMolecule> standardizationResults = chemStructureService
-				.standardizeStructures(parentIdsToStructures);
```
brianbolt, line 717: This is the good part of removing the call to standardize the parents AGAIN unnecessarily and instead just fetch the cmpdRegmolecules (703)

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -764,23 +749,27 @@ public int restandardizeParentStructuresBatch(List<Long> pIdGroup) throws CmpdRe
 
 			parent.setMolFormula(cmpdRegMolecule.getFormula());
 
-			parent.merge();
```
brianbolt, line 767: You don't actually need to call merge, this is all handled by hibernate and it slows things down.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -817,12 +807,10 @@ public String executeStandardization(String username, String reason) throws Stan
 		standardizationHistory.setStandardizationUser(username);
 		standardizationHistory.setStandardizationReason(reason);
 		standardizationHistory.merge();
-		this.reset();
```
brianbolt, line 820: We reset these tables when doing dry run and I feel like its not necessary to reset them after doing standardization because they can be used to interrogate change later (at least it was helpful to me).

bffrost (commented):
This is pretty complex, and a little hard for me to fully understand the implications of.
One question about the strategy shift from "rerun full standardization when you go to apply" to "apply dry run results directly": Have you considered the case when an admin runs a dryrun standardization, some time elapses and people register more compounds, then the admin applies the dryrun standardization? It seems like we have the possibility of some "extra" compounds that won't get restandardized.

If we decide to worry about that, it might make sense to retain some of the old logic, but only use it for the parents that aren't covered by the dryrun table. Does that make sense?

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -606,7 +601,7 @@ public static TypedQuery<StandardizationDryRunCompound> findStandardizationDryRu
 			throw new IllegalArgumentException("The corpName argument is required");
 		EntityManager em = StandardizationDryRunCompound.entityManager();
 		TypedQuery<StandardizationDryRunCompound> q = em.createQuery(
-				"SELECT o FROM StandardizationDryRunCompound AS o WHERE o.corpName = :corpName",
+				"SELECT o FROM StandardizationDryRunCompound AS o JOIN o.parent p WHERE p.corpName = :corpName",
```
bffrost, line 604: nit: Should this use `JOIN FETCH` ?

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -618,7 +613,7 @@ public static TypedQuery<StandardizationDryRunCompound> findStandardizationDryRu
 			throw new IllegalArgumentException("The corpName argument is required");
 		EntityManager em = StandardizationDryRunCompound.entityManager();
 		StringBuilder queryBuilder = new StringBuilder(
-				"SELECT o FROM StandardizationDryRunCompound AS o WHERE o.corpName = :corpName");
+				"SELECT o FROM StandardizationDryRunCompound AS o JOIN o.parent p WHERE p.corpName = :corpName");
```
bffrost, line 616: nit: Should this use `JOIN FETCH` ?

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -645,18 +640,19 @@ public static TypedQuery<StandardizationDryRunCompound> checkForDuplicateStandar
...
 		
 		EntityManager em = StandardizationDryRunCompound.entityManager();
-		String queryBuilder = "SELECT o FROM StandardizationDryRunCompound AS o WHERE o.CdId IN (:cdIds) AND o.id != :id and o.stereoCategory = :stereoCategory";
+		String queryBuilder = "SELECT o FROM StandardizationDryRunCompound AS o JOIN FETCH o.parent p WHERE o.CdId IN (:cdIds) AND o.id != :id and p.stereoCategory = :stereoCategory";
 
-		if(standardizationDryRunCompound.getParent().getStereoComment() == null) {
-			queryBuilder += " AND o.stereoComment IS NULL";
+		Boolean stereoCommentEmpty = standardizationDryRunCompound.getParent().getStereoComment() == null || standardizationDryRunCompound.getParent().getStereoComment().length() == 0;
+		if(stereoCommentEmpty) {
+            queryBuilder += " AND (p.stereoComment IS NULL or p.stereoComment = '')";
```
bffrost, line 647: nit: funny indentation

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -600,11 +617,11 @@ public String writeStandardizationCompoundsToSDF(List<StandardizationDryRunCompo
 						String.valueOf(stndznCompound.isAsDrawnDisplayChange()));
 				cmpdRegMolecule.setProperty("Stereo Category", stndznCompound.getParent().getStereoCategory().getName());
 				cmpdRegMolecule.setProperty("Stereo Comment", stndznCompound.getParent().getStereoComment());
-				sdfWriter.writeMol(cmpdRegMolecule);
+                cmpdRegMoleculeList.add(cmpdRegMolecule);
```
bffrost, line 620: nit: weird indent here

bffrost (approved):
LGTM, all the new changes since my last review make sense, only found nits with spacing.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
brianbolt: It actually really depends on how you are using it.  JOIN FETCH can speed things up but it's not strictly necessary and isn't obvious that its needed in this context.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
brianbolt: It actually really depends on how you are using it. JOIN FETCH can speed things up but it's not strictly necessary and isn't obvious that its needed in this context.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Thanks, fixed.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
brianbolt: Thanks, fixed.

brianbolt:
I snuck in one more change to speed up the download of SDF files. Going to merge it because I tested and it is significantly faster for multi-thousand downloads.

PR #380, opened 2022-10-19
ACAS-488: Write list of mols to SDF

brianbolt:
## Description
 - Add `writeMols(List<CmpdRegMolecule> cmpdRegMoleculeList)`
 - This speeds up some implementations which are faster writing to SDF in bulk than one at a time
 - Fixed a bug that we weren't closing the file which is required to fully flush the file writer. Without this change we could intermittently loose a few mols in the output file.

## Related Issue
ACAS-488

## How Has This Been Tested?
Tested by downloading 18K compounds which takes about 1 minute with this new code


bffrost (approved):
LGTM

PR #383, opened 2022-10-21
ACAS-495: Fix salt_form fill structures

brianbolt:
## Description
 - Fix issue with calling the wrong method and not actually updating salt_form structures
 - Fix comment which referenced parents instead of salt_forms

## Related Issue
ACAS-495

## How Has This Been Tested?
Truncated a bbchem_salt_form_structure table and restarted ACAS
Verified bbchem_salt_form_structure was re populated with all structures from salt_form where mol_structure was not empty.
Verified that salt_form.cd_ids were update with corresponding new entries.


bffrost (approved):
LGTM

PR #385, opened 2022-10-21
ACAS-473: Fix BBChem in-memory SDF writer

bffrost:
## Description
- This change is a fix for the in-memory / file-less SDF writer in the BBChem implementation. It is used by the SaltService to produce an SDF when users download all salts.
- Prior implementation returned a string like `java.io.FileWriter@490b189e`
- This fixes it to use a StringWriter whose "toString" method works as intended and returns all the text that has been written to the writer.

## How Has This Been Tested?
Tested on frost-acas-quickstart.dev.bb.schrodinger.com by tagging `ACAS-473-dev1` from my dev branch, and deploying the tag onto that server which uses bbchem.
Confirmed the issue before updating roo, and confirmed it produces a proper SDF after this change.

### DIFF
--- a/src/main/java/com/labsynch/labseer/chemclasses/bbchem/CmpdRegSDFWriterBBChemImpl.java
+++ b/src/main/java/com/labsynch/labseer/chemclasses/bbchem/CmpdRegSDFWriterBBChemImpl.java
@@ -1,8 +1,9 @@
 package com.labsynch.labseer.chemclasses.bbchem;
 
-import java.io.File;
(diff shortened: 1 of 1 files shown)


brianbolt (approved):
LGTM. Looking through the code I think this previously slipped through the cracks because the only place that called getCmpdRegSDFBufferWriter (convertMolfilesToSDFile) wasn't actually being used anywhere so never fully tested it when implementing bbchem.

PR #386, opened 2022-10-24
Roo j chem test upgrade

philippcheung:
Roo Java 11/JChem 22 upgrade

## Description
This pull request contains both an upgrade to Java 11 and an upgrade to JChem 22

## How Has This Been Tested?
This has passed all ACASClientPy Unit tests.


Dockerfile-multistage-j11
```diff
@@ -0,0 +1,88 @@
...
+FROM 	builder as dependencies
+ARG     CHEMISTRY_PACKAGE
+ENV     CHEMISTRY_PACKAGE=${CHEMISTRY_PACKAGE}
+
+RUN pwd
+FROM 	dependencies as jchem
+ADD settings.xml /root/.m2/settings.xml
+
+# ADD 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
+# RUN     mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
```
brianbolt, line 13: We shouldn't include this jar any more if we are building using newer jchem right?  I think you should remove any commented out lines but this will be more noticeable once you replace `Dockerfile-multistage`

pom.xml
```diff
@@ -33,17 +33,17 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.]]>
...
 		<maven.build.timestamp.format>yyyyMMdd_HHmm</maven.build.timestamp.format>
 		<date>${maven.build.timestamp}</date>
 		<slf4j.version>1.7.5</slf4j.version>
 		<log4j.version>2.17.2</log4j.version>
 		<spring.version>5.3.19</spring.version>
 		<querydsl.version>3.4.3</querydsl.version>
+
 		<spring-security.version>5.6.3</spring-security.version>
 		<cdk.version>1.5.13</cdk.version>
-		<jchem.version>16.4.25.0</jchem.version>
```
brianbolt, line 46: Should we have just changed this to 2022.12.0?  Seems like we should continue using this properties section and then use the `jchem.version` variable below.

src/main/java/com/labsynch/labseer/chemclasses/jchem/ChemStructureServiceJChemImpl.java
```diff
@@ -66,6 +66,9 @@
 import org.apache.commons.io.FileUtils;
 import java.util.UUID;
 
+
+
+
```
brianbolt, line 71: *nit - remove extra white space

src/main/resources/db/migration/jchem/postgres/V2022_1_X__Add_jchem_migration.java
```diff
@@ -0,0 +1,54 @@
...
+import chemaxon.jchem.db.Updater;
+
+public class V1_3_0_2__Add_standardization_jchem_tables implements JdbcMigration {
+
+	Logger logger = LoggerFactory.getLogger(V1_3_0_2__Add_standardization_jchem_tables.class);
+
+	// Updating jchem version
+
+	public void migrate(Connection conn) throws Exception {
+		logger.info("creating standardization_dry_run_structure table");
```
brianbolt, line 26: This comment doesn't match what the code does

README.md
```diff
@@ -27,4 +27,10 @@ ACAS-Roo Server
...
 
 * Repo owner or admin
-* Other community or team contact
+* Other community or team contact
+
+
+## JChem Changes
+
+* For JChem and Java 11, there are new changes that use the JChem Maven repository for the build process
+* This requires additional authentication changes that reference ~/.m2/settings.xml
```
brianbolt, line 36: I think we need more detail. Can we provide a settings.xml for use here? Wondering also if it's possible to keep the old build process.  I know we have a few automated builds on servers which could benefit from using the old big fat jar method.  i.e. can we make the Dockerfile-jchem file and Dockerfile-multistage file still use a jchem jar if it's provided rather than the settings.xml?

Dockerfile-multistage
```diff
@@ -1,68 +1,52 @@
...
-RUN 	--mount=type=cache,target=/root/.m2 mvn clean && \
+RUN   mvn clean && \
+# RUN 	--mount=type=cache,target=/root/.m2 mvn clean && \
         mvn compile war:war -P ${CHEMISTRY_PACKAGE}
+RUN mvn dependency:resolve
+FROM tomcat:9.0.62-jre11-openjdk-slim-buster
 
-FROM tomcat:9.0.62-jre8-openjdk-slim-buster
-
-# Third and Last Step That Requires Significant (Relative) Amount of Time
```
brianbolt, line 29: Can you restore this comment?

Dockerfile-multistage
```diff
@@ -1,68 +1,52 @@
...
+# RUN 	--mount=type=cache,target=/root/.m2 mvn clean && \
         mvn compile war:war -P ${CHEMISTRY_PACKAGE}
+RUN mvn dependency:resolve
+FROM tomcat:9.0.62-jre11-openjdk-slim-buster
 
-FROM tomcat:9.0.62-jre8-openjdk-slim-buster
-
-# Third and Last Step That Requires Significant (Relative) Amount of Time 
 RUN apt-get update && \
-    apt-get install -y \
```
brianbolt, line 31: We have these all on new lines to make reviews easier when we change a package version. If there wasn't a reason to change this, can you restore it?

Dockerfile-multistage
```diff
@@ -1,68 +1,52 @@
...
-      libfreetype6 \
-      libfontconfig \
-      openssl 
+    apt-get install -y openssl libfontconfig libfreetype6 curl
 
 # Add nodejs for prepare config files
 ENV NPM_CONFIG_LOGLEVEL warn
 ENV NODE_VERSION 14.x
-
-# Second Slowest Step
```
brianbolt, line 41: Can your restore this comment if there was no reason to change it?

Dockerfile-multistage
```diff
@@ -1,68 +1,52 @@
...
+    apt-get install -y openssl libfontconfig libfreetype6 curl
 
 # Add nodejs for prepare config files
 ENV NPM_CONFIG_LOGLEVEL warn
 ENV NODE_VERSION 14.x
-
-# Second Slowest Step 
 RUN curl -fsSL https://deb.nodesource.com/setup_$NODE_VERSION | bash - && \
   apt-get install -y nodejs && \
-  npm install -g \
```
brianbolt, line 44: This changed some package versions and formatted it differently. Can you restore the original version if there was no reason to change it?

Dockerfile-multistage
```diff
@@ -1,68 +1,52 @@
...
 
 FROM 	${CHEMISTRY_PACKAGE} as compile
-
-COPY 	pom.xml /src/pom.xml
+ADD 	pom.xml /src/pom.xml
 WORKDIR /src
-RUN 	--mount=type=cache,target=/root/.m2 mvn dependency:resolve -P ${CHEMISTRY_PACKAGE}
+RUN 	mvn dependency:resolve -P ${CHEMISTRY_PACKAGE}
+# Caching doesn't seem to work with the ChemAxon maven repo
+# RUN 	--mount=type=cache,target=/root/.m2 mvn dependency:resolve -P ${CHEMISTRY_PACKAGE}
```
brianbolt, line 20: This file generally worked before and definitely sped up the use case of speeding up indigo and bbchem chemistry packages before.  We need to maintain that speed.  Can we ever these cache changes? 

There is a separate file that needs to get updated called Dockerfile-jchem.  I think you should update that file with the equivalent changes made for jchem in this file, and then just use the Dockerfile-jchem to build the docker image.

brianbolt (changes requested):
The code changes look good to me.  I added some notes on things to revert as I think they were mistakenly changed. We also need to maintain the cache mechanism as it sped up our builds significantly for indigo and bbchem. It seems like jchem builds were broken for Dockerfile-multistage before you started this work so I think the easiest thing to do is port your changes for Jchem over to Dockerfile-jchem and build the images with it instead of Dockerfile-multistage.

Dockerfile-multistage
```diff
@@ -1,68 +1,53 @@
...
 ADD 	. /src
 
 # Slow Step - Caching this Layer Greatly Improves Build Time  (~6x Faster) 
-RUN 	--mount=type=cache,target=/root/.m2 mvn clean && \
+RUN   mvn clean && \
+# RUN 	--mount=type=cache,target=/root/.m2 mvn clean && \
         mvn compile war:war -P ${CHEMISTRY_PACKAGE}
-
-FROM tomcat:9.0.62-jre8-openjdk-slim-buster
+RUN mvn dependency:resolve
```
brianbolt, line 27: Why run dependency resolve at the end?

pom.xml
```diff
@@ -251,6 +254,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.]]>
 			<artifactId>commons-io</artifactId>
 			<version>2.11.0</version>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-text</artifactId>
+			<version>1.10.0</version>
+		</dependency>
```
brianbolt, line 261: Can you describe why this was added?

Dockerfile-multistage
```diff
@@ -1,68 +1,53 @@
...
+ADD settings.xml /root/.m2/settings.xml
 
 FROM 	dependencies as bbchem
 
 FROM 	dependencies as indigo
 
 FROM 	${CHEMISTRY_PACKAGE} as compile
-
-COPY 	pom.xml /src/pom.xml
+ADD 	pom.xml /src/pom.xml
```
brianbolt, line 16: Why did this get changed from COPY to ADD?

brianbolt (changes requested):

brianbolt:
@bffrost This should be ready for review.  We removed the extra docker files and updated Dockerfile multistage to work for all 3 chemistry engines and verified they build correct on Phil's machine.

pom.xml
brianbolt: @jmcneil86043 Correct me if I'm wrong here but Phil mentioned that you all noticed that even release/2022.4.x was affected by this?  I wonder if this is a dependency that Jchem pulls in that only affects jchem builds of acas-roo-server.  @philippcheung One way you can check this is by running `man dependency:tree` and viewing the output for both indigo and jchem (without this change).  My bet is that you see `org.apache.commons` in the jchem build and not the indigo build.  If that is the case, we should move this dependency to inside the jchem profile of the pom.xml

philippcheung: Just to clarify what I meant, and this might be in agreement with what you're hypothesizing -- my new version of the roo (before I added this) caused a security trigger, while stage machines didn't trigger the security alert -- so yes, it might be the jchem.

brianbolt: Yeah, so I removed the fix above and I see this in the depdendency tree:

```
[INFO] +- com.chemaxon:jchem-main:jar:22.12.0:compile
...etc.
[INFO] |  +- com.chemaxon:descriptors-jep:jar:22.12.0:compile
[INFO] |  |  |  +- com.chemaxon:io-all:jar:22.12.0:compile
[INFO] |  |  |  |  +- com.chemaxon:io-csv:jar:22.12.0:compile
[INFO] |  |  |  |  |  \- org.apache.commons:commons-text:jar:1.9:compile
```
 Which means that the issue is coming from the jchem jar.  I just moved the commons text under the jchem profile so it only gets pulled into the build when doing a jchem build.

bffrost (approved):
LGTM. Changes are narrow and should only affect JChem builds. Migration has a unique name and doesn't clash with any migrations on 2023.1.x branch.

I appreciate the notes in the pom.xml about the jchem-specific dependencies.

Ship it!

brianbolt:
@bffrost I think you know this but I just wanted to make it crystal clear that this does affect non Jchem builds as it bumps us from java 8 to java 11

PR #387, opened 2022-10-25
ACAS-387: LD Chem config check, config fix and health check

brianbolt:
## Description
  -  For BBChem implementation refactored to just a single URL with paths built into the code
  - `checkStandardizerSettings` chemstructure service interface and implementations to pass the applied standardization settings and system configured standardization settings and determine if standardization is needed
    - Keep the jchem service implementation the same (just compare hashes)
    - No check for indigo as standardization is not implmented
    - For BBChem
      - Call service to fix configs (check they are valid)
      - Call service to compare old (fixed configs) and new (fixed configs)
         - Add the Preprocessor and Suite version to the saved settings to be able to pass to the config check service.  Did this by adding a call to the health endpoint.
 - Refactor image service function to be in BBChemservice controller

## Related Issue
ACAS-387

## How Has This Been Tested?
Build all 3 indigo, jchem and bbchem implementations


src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
```diff
@@ -48,35 +47,31 @@ public class BBChemStructureServiceImpl implements BBChemStructureService {
...
-		JsonNode urlNode = jsonNode.get(propertyKey);
-		if (urlNode == null || urlNode.isNull()) {
-			logger.error("Missing preprocessorSettings " + propertyKey + "!!");
-			throw new IOException("Missing preprocessorSettings " + propertyKey + "!!");
-		}
-		String url = urlNode.asText();
-		return url;
-
-	}
-
```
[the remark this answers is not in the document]
bffrost: That's great. We'll want to update the default k8s configs and the config upgrader script once we merge this, right?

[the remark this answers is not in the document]
brianbolt: Yes, thanks, definitely need to set it on K8s values. I JUST sent you a seperate PR for this.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -90,104 +109,110 @@ public void onApplicationEvent(ContextRefreshedEvent event) {
...
+		// Get the applied settings from the history table which should have the most recent standardization settings applied
+		List<StandardizationHistory> completedHistories = StandardizationHistory.findStandardizationHistoriesByStatus("complete", 0, 1, "id", "DESC").getResultList();
+
+		StandardizationHistory mostRecentHistory;
+		if (completedHistories.size() > 0) {
+			// Got a complete history where the standardization actually standaradized compounds
+			mostRecentHistory = completedHistories.get(0);
+		} else {
+			// There are 2 normal states where there are no completed histories
+			// 1. We upgraded to the new standardization code and there are no completed standardizations
```
bffrost, line 145: This comment block is useful, but I think it should be expanded a bit.
Specifically, explaining what will happen in this case 1, since it's not part of the `if` statement below. I'm thinking something like:
```
// In case 1, we pass along an empty StandardizationHistory, which will cause the system to report that it needs standardization.
// In case 2 we try to fill in an initial standardization record with the current settings, provided they're valid settings, so no restandardization is needed.
```

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -90,104 +109,110 @@ public void onApplicationEvent(ContextRefreshedEvent event) {
...
 			}
-		} catch (StandardizerException e) {
-			logger.error("Caught error on checking standardization state", e);
+
+		}
+
+		// Do a config check to verify that settings are valid and if we need to restandardize or not
+		StandardizationSettingsConfigCheckResponseDTO configCheckResponse = chemStructureService.checkStandardizerSettings(mostRecentHistory, currentRawStandardizerSettings);
+
+		// We should only ever really have one standardization settings row at any given
```
bffrost, line 176: I think it is worth explaining here what a "StandardizationSettings" is. A lot of these variables and classes have similar names. As I understand it `StandardizationSettings` is actually the current state of the system, specifically whether it needs restandardization and why. `StandardizationStatus` might be a better name, or `CurrentStandardizationStatus`

bffrost (changes requested):
Looks good. This is a large set of changes but I'm able to follow it, and the refactor has gone a long way to simplifying this stuff. Spotted one attribute that I think is unused, and left some comments about code comments and a confusing class name. Should be good to go once these are addressed.

src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
```diff
@@ -679,7 +659,7 @@ public HashMap<? extends AbstractBBChemStructure, Boolean> substructureMatch(Str
 		for (AbstractBBChemStructure needsMatch : needsMatchStructures) {
 			arrayNode.add(needsMatch.getMol());
 		}
-		requestData.put("needs_match", arrayNode);
+		requestData.replace("needs_match", arrayNode);
```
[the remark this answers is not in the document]
brianbolt: Deprecation notice: https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.13.2.2/com/fasterxml/jackson/databind/node/ObjectNode.html#put-java.lang.String-com.fasterxml.jackson.databind.JsonNode-

Says to use set or replace:
https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.13.2.2/com/fasterxml/jackson/databind/node/ObjectNode.html#set-java.lang.String-com.fasterxml.jackson.databind.JsonNode-
https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.13.2.2/com/fasterxml/jackson/databind/node/ObjectNode.html#replace-java.lang.String-com.fasterxml.jackson.databind.JsonNode-

Only difference is that set returns the node and replace returns the old value but we aren't using the return values in either case.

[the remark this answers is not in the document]
brianbolt: I didn't replace all deprecated uses but I will do that now.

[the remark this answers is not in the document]
brianbolt: Done

src/main/java/com/labsynch/labseer/domain/StandardizationHistory.java
```diff
@@ -75,6 +76,8 @@ public class StandardizationHistory {
 
     private Integer registrationErrorCount;
 
+    private String needsStandardiationReasons;
```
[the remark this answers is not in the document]
brianbolt: Thanks this was a mistake, didn't mean to add it to this class. Good catch.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Thanks, added those

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Makes sense but unclear if you are suggesting that I refactor the table name, class name and variable names to match `CurrentStandardizationStatus ` as part of this PR or to just note it in this comment.

brianbolt: I added some comments at least for now.

brianbolt: I realize after writing this that the way to refactor this is to just drop the settings table and make settings just a DTO because there is no point in saving the row in a table if we are completely replacing it on every check.

bffrost: @brianbolt your last comment makes a lot of sense to me. I think it's worth doing as part of this PR, just to leave this code in a clean place. My hope is that we don't have a lot more work to do in this area of the code, so there may not be another good opportunity to make that change.

bffrost (approved):
LGTM, nice simplification in your latest refactor.

PR #389, opened 2022-10-28
ACAS-502: Max auto lot number for bulk loader

brianbolt:
## Description
 - Use max auto lot number if set to select lot numbers < or equal to the max auto lot number when getting the next lot number during bulk reg and reparent lot
 - Previously reparent lot paid attention to maxAutolot number only for the purposes of keeping a lot number if it was above the max auto lot number, however if the target parent had a lot greater than the max auto auto lot number it would still grab this large lot number and add 1 to it.

## Related Issue
ACAS-502

## How Has This Been Tested?
Loaded a compound with lot number 1
Added a lot manually with lot number 1001
Bulk loaded another lot and verified the lot number assigned was lot 2
Reparented a lot onto a parent that had a lot number of 1001 and verified it was assigned lot 3 (lowest lot number below the max auto lot number)

### DIFF
--- a/src/main/java/com/labsynch/labseer/domain/Lot.java
+++ b/src/main/java/com/labsynch/labseer/domain/Lot.java
@@ -381,7 +381,7 @@ public class Lot {
         return q.getSingleResult();
     }
 
-    public static Integer getMaxParentLotNumber(Parent parent) {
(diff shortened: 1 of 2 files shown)


bffrost (approved):
LGTM! Glad this was such a small change

PR #391, opened 2022-11-02
Fixed: Alias dropped when new lot of compound gets registered in bulk loader

dalejerikson:
**Steps to Reproduce**

Bulk load a new structure with a Parent Alias mapped in. This creates a new parent with the alias, and a new lot.

Then load the same structure with a different Parent Alias value. This will create a new lot on the existing parent.

**Bug Behavior**

A new lot is created, but the alias from the second upload is ignored. The parent ends up with one alias (only the one from the first upload).

**Fixed Behavior**

A new lot will be created and the existing parent will be updated to add the new Parent Alias value. The parent will end up with two aliases (one from each upload)

**Example**

One of the tests I conducted was to have one compound with the alias "Dale's Favorite Compound" and I uploaded the same structure with the alias of "Second Alias". The result was two lots with the same parent. The parent contains both aliases. 

Screen Shot 2022-11-02 at 11 09 04 AM

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -837,15 +837,13 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 
 						Set<String> parentAliasStrings = new HashSet<>();
 						for(ParentAlias alias : parent.getParentAliases()){
-							if (! alias.isIgnored())
-							{
+							if (! alias.isIgnored()) {
 								parentAliasStrings.add(alias.getAliasName());
 							}
 						}
 						Set<String> foundParentAliasStrings = new HashSet<>(); 
 						for(ParentAlias alias : foundParent.getParentAliases()){
-							if (! alias.isIgnored())
-							{
+							if (! alias.isIgnored()) {
 							foundParentAliasStrings.add(alias.getAliasName());
 							}
 						}
@@ -863,8 +861,7 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 						// Only need to do this if no problems found in previous loop; otherwise redundant 
 						// to processs that occurs after
 						{
-							for(ParentAlias alias : foundParent.getParentAliases())
-							{
+							for(ParentAlias alias : foundParent.getParentAliases()) {
 								// Check alias in parentAlias (String Comparison)
 								// if not found then equal Aliases is false 
 								if(! alias.isIgnored() && !parentAliasStrings.contains(alias.getAliasName())){
@@ -874,25 +871,20 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 						}	
 
 						// If Incoming Parent and Found Parent Have Equal Aliases 
-						if(equalAliases)
-						{
+						if(equalAliases) {
 							// No Changes or Modifications Needed; Can Break Loop
 							// Continue 
 							parent = foundParent;
 							break searchResultLoop; 
-						}
-						else
-						{
+						} else { 
 							Set<ParentAlias> unionParentAliases = parent.getParentAliases();
 							// Parent Would Be Found Parent w/ Appropriate Updates
 							parent = foundParent;
 							// Update Parent Object to Have "Union" of All Aliases 
-							for(ParentAlias oldAlias : parent.getParentAliases())
-							{
+							for(ParentAlias oldAlias : parent.getParentAliases()) {
 								// If oldAlias Not In UnionParentAliases (Use Previously Set<str> to Do Compare)
 									// Add to Aliases Union List 
-								if(! oldAlias.isIgnored() && !parentAliasStrings.contains(oldAlias.getAliasName()))
-								{
+								if(! oldAlias.isIgnored() && !parentAliasStrings.contains(oldAlias.getAliasName())) {
 									unionParentAliases.add(oldAlias);
 								}
 							}
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -826,50 +826,7 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 					if (sameStereoCategory & sameStereoComment
 							& (sameCorpName | (noCorpName & sameCorpPrefixOrNoPrefix))) {
 						// parents match (based on above criteria)
-						boolean equalAliases = true; // assumption is aliases are same unless proven to be false
-
-						// Need to Create String Sets to Do 
-						// "Is alias (str) of parent in aliases of foundParent (set of strs)?" (and vice versa) logic 
-
-						// Note: This is not the most efficient / concise code to do this process
-						// however due to readability and the assumingly small number of aliases 
-						// parents will have this is fine 
-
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -825,9 +825,77 @@ public Parent validateParent(Parent parent, Collection<BulkLoadPropertyMappingDT
...
+						// however due to readability and the assumingly small number of aliases 
+						// parents will have this is fine 
+
+						Set<String> parentAliasStrings = new HashSet<>();
+						for(ParentAlias alias : parent.getParentAliases()){
+							parentAliasStrings.add(alias.toString());
+						}
+						Set<String> foundParentAliasStrings = new HashSet<>(); 
+						for(ParentAlias alias : foundParent.getParentAliases()){
+							foundParentAliasStrings.add(alias.toString());
```
bffrost, line 844: What does this `alias.toString()` method return? Is it just the `aliasName`? Or is it an Object like `ParentAlias[lsType=null,lsKind=null,id=27,aliasName="my favorite alias"]`?

bffrost:
@dalejerikson can you add some acasclient tests for this new behavior? Specifically some confirming the "negative" cases of loading the same alias or aliases twice and confirming we don't create duplicate aliases.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -825,9 +825,77 @@ public Parent validateParent(Parent parent, Collection<BulkLoadPropertyMappingDT
 							|| foundParent.getCorpName().contains(parent.getLabelPrefix().getLabelPrefix()));
 					if (sameStereoCategory & sameStereoComment
 							& (sameCorpName | (noCorpName & sameCorpPrefixOrNoPrefix))) {
-						// parents match
-						parent = foundParent;
-						break searchResultLoop;
+						// parents match (based on above criteria)
+						boolean equalAliases = true; // assumption is aliases are same unless proven to be false
+
+						// Need to Create String Sets to Do
```
bffrost, line 831: nit: Just because a bunch of the functions in this `BulkLoadServiceImpl` are already quite long, It'd be nice to encapsulate this into a helper function like `private Parent checkAndUpdateParentAliases(Parent parent, Parent foundParent)`. That'll help maintain readability in this already long `validateParent` function.

bffrost (changes requested):
@dalejerikson the general flow of the logic looks good, but I'm concerned that the `toString()` method may not work the way you expect it does. The best way to confirm this would be to write a few acasclient tests and post them for PR.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
dalejerikson: I suppose `getAlias()` would be more appropriate. Also I forgot to add in if it is marked as ignored so I will do that as well.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
dalejerikson: That's fair.

dalejerikson:
Interesting: When adding tests to acas client, I noticed that there is already something that tests manipulating parent aliases.

https://github.com/mcneilco/acasclient/blob/release/2022.4.x/tests/test_acasclient.py#L3081 

Before this fix, it theoretically should have failed. Looks like the logic it different and more directly interacts with the Parent Alias object than the bug resulting from the issue in the bulk loader that this ticket fixes. 

Will request re-review for this now but i'll look to add tests that more accurately reflect this ticket.

dalejerikson:
ACAS Client Test PR: https://github.com/mcneilco/acasclient/pull/101

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
brianbolt: @bffrost Comment was marked as resolved but I don't think it actually was, right?

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -825,9 +825,83 @@ public Parent validateParent(Parent parent, Collection<BulkLoadPropertyMappingDT
...
+						// "Is alias (str) of parent in aliases of foundParent (set of strs)?" (and vice versa) logic 
+
+						// Note: This is not the most efficient / concise code to do this process
+						// however due to readability and the assumingly small number of aliases 
+						// parents will have this is fine 
+
+						Set<String> parentAliasStrings = new HashSet<>();
+						for(ParentAlias alias : parent.getParentAliases()){
+							if (! alias.isIgnored())
+							{
```
brianbolt, line 841: nit - Can we stick with { being on the same line as the if or else block since it's that way in most of the code?

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -825,9 +825,83 @@ public Parent validateParent(Parent parent, Collection<BulkLoadPropertyMappingDT
...
+
+						// If Incoming Parent and Found Parent Have Equal Aliases 
+						if(equalAliases)
+						{
+							// No Changes or Modifications Needed; Can Break Loop
+							// Continue 
+							parent = foundParent;
+							break searchResultLoop; 
+						}
+						else
```
brianbolt, line 884: nit - Same nit as above with keeping else on the same line as } and {

bffrost (approved):
LGTM! The fixes for string comparison look good.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
bffrost: Good catch @brianbolt . @dalejerikson  Can you look at pulling this logic out into a helper function just to improve readability?

dalejerikson: I left an additional separate comment about the break statement that is in the code

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -825,9 +825,75 @@ public Parent validateParent(Parent parent, Collection<BulkLoadPropertyMappingDT
...
+								}
+							}
+						}	
+
+						// If Incoming Parent and Found Parent Have Equal Aliases 
+						if(equalAliases) {
+							// No Changes or Modifications Needed; Can Break Loop
+							// Continue 
+							parent = foundParent;
+							break searchResultLoop;
```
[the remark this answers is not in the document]
bffrost: What about two helpers? `compareParentAliases` and `updateParentAliases`
Then you can keep the `if(equalAliases)` block and the `break` statement, but still keep the complexity down.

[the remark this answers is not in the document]
dalejerikson: Implemented

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
dalejerikson: Implemented

PR #394, opened 2022-11-05
ACAS-413: DB Migration for ACAS-304 endpoint management

bffrost:
## Description
- Fairly involved database migration to bring older data up to new current state for ACAS-304
- Add all column names, column units, and concentration units into Data Dictionaries
- Create column order states for very old experiments that are missing them
- Create data column order states on protocols, as the union of all columns on their experiments
- Switch several 'stringValue' fields within column order states over to 'codeValue'

## How Has This Been Tested?
Local environment testing
- Switched to release/2022.4.x branch (which does not have R changes from ACAS-411 to save protocol column order states), loaded several experiments, then applied the migration and switched back to ACAS-447 (recent dev branch from @hhan-schrodinger ) and inspected experiments and protocols in the UI
- Deleted data column order states and values from a few experiments, ran the migration to make sure they got populated.

This really could use more testing. @hhan-schrodinger can you try switching to this Roo branch and running it on your dev environment to see if it successfully migrates everything you've got loaded locally?

We'll also want to run this on some older staging servers (or clones of them) and spot check older protocols and experiments in the UIs.


hhan-schrodinger:
@bffrost Tried it on my local environment and everything is working and was converted successfully!

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
```diff
@@ -0,0 +1,338 @@
...
+
+--column name
+select add_data_column_order_ddict_value(column_name, 'column name'::varchar) from
+(select distinct column_name from tmp_expt_columns) a;
+--units
+select add_data_column_order_ddict_value(units, 'column units'::varchar) from
+(select distinct units from tmp_expt_columns) a where a.units is not null;
+--conc units
+select add_data_column_order_ddict_value(conc_units, 'column conc units'::varchar) from
+(select distinct conc_units from tmp_expt_columns) a where a.conc_units is not null;
```
hhan-schrodinger, line 53: Is there a reason only column name, units, and conc units are being filled in here?

hhan-schrodinger (approved):
Looks good to me - I really like the step by step explanations and sections in the code comments. Helped a lot with following along and learning the process!

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
bffrost: Are there other data dictionaries that we need to fill in?
"column type" is a fixed set of options that'll be filled in by the ConfJSON. Concentration and Time are numeric (no data dictionary). I did skip "time units" because I don't think it's really being used.

hhan-schrodinger: Got it, that makes sense! Time units was the only one that I could think of.

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
```diff
@@ -0,0 +1,338 @@
...
+
+-- Build a temp table of all experiment data column order stuff
+create table tmp_expt_column_order_data as
+   -- Use row_number() to reassign column order at the protocol level, since different experiments may have different columns in different order
+   -- The 'earliest experiment wins', meaning any new endpoints from later experiments will be added later in the column order
+	select *, row_number() over (partition by protocol_id order by experiment_id, column_order) as new_column_order
+	from (
+		-- This query pivots all the 'data column order' experiment values into a table of column_type, column_name, etc.
+		select e.protocol_id,
+			coalesce(col_type.code_value, col_type.string_value) as column_type, --the coalesce is used to account for having the same data in stringValues or in codeValues
```
brianbolt, line 152: The PR description I think mentions this as:
> Switch several 'stringValue' fields within column order states over to 'codeValue'

But I'm not sure what this is. Is this to fix some local development servers because I don't see where string values exist in this state.

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
```diff
@@ -0,0 +1,338 @@
...
+			coalesce(col_name.code_value, col_name.string_value),
+			col_conc.numeric_value,
+		 	coalesce(col_conc_units.code_value, col_conc_units.string_value),
+			coalesce(hide_col.code_value, hide_col.string_value),
+			coalesce(cond_col.code_value, cond_col.string_value)
+		) a;
+
+-- Define a function to create the protocol column order states and values
+-- this is nearly identical to the function above, just using protocol instead of experiment
+create or replace function create_protocol_data_column_order_state(protocol_id bigint, transaction_id bigint, column_order int, column_type varchar, column_name varchar,
```
brianbolt, line 186: Wondering if before this, we should fill in a `"strict endpoint matching"` value into the protocol metadata state and default it to `false`

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
```diff
@@ -0,0 +1,338 @@
...
+									  ls_kind, ls_transaction, ls_type, ls_type_and_kind, public_data, recorded_by, recorded_date, 
+									  version, experiment_state_id)
+			select nextval('value_pkseq'), 'boolean', 'ACAS DDICT', 'boolean', 'boolean_boolean', case when hidden then 'TRUE' else 'FALSE' end, false, false,
+									  'hide column', transaction_id, 'codeValue', 'codeValue_hide column', true, 'ACAS', now(),
+									  0, expt_state_id;
+		-- create condition value
+		insert into experiment_value (id, code_kind, code_origin, code_type, code_type_and_kind, code_value, deleted, ignored, 
+									  ls_kind, ls_transaction, ls_type, ls_type_and_kind, public_data, recorded_by, recorded_date, 
+									  version, experiment_state_id)
+			select nextval('value_pkseq'), 'boolean', 'ACAS DDICT', 'boolean', 'boolean_boolean', case when condition then 'TRUE' else 'FALSE' end, false, false,
```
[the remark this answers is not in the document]
brianbolt: This would obviously have ramifications for the ACAS and R code.  I'd have to check but I believe lines like this get json converted into lowercase `true` values.

https://github.com/mcneilco/acas/blob/38675b89ff98d7a004a8f515bd5c4bf9b56c9fcf/modules/GenericDataParser/src/server/generic_data_parser.R#L2043

brianbolt (changes requested):
Overall makes sense but I left some notes in the review which should be looked at like casing of `TRUE` vs `true` and using p_api view during the migration.

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
brianbolt: Ah, now I see https://github.com/mcneilco/acas/pull/1033

But still curious as to why we added a migration for this for a value which isn't in production. I might still be missing something here.

bffrost: We do need to fix the stringValues on Experiments since they already have these states.
For protocols, yeah, those values are not in production. I added it to cover Hansen's dev environment, and because it was relatively simple to add.

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
bffrost: Yeah, I think you're right. I'll add that.

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
bffrost: I see what you mean now - the coalesce isn't necessary. I'll remove those to simplify it.

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
bffrost: Actually, I'm not going to. I think it's more accurate to be missing a value, and program the R & frontend code to use the default from the configs in cases where protocols are missing that value. If I bulk populate everything with `false` then it'll make setting the default strictMatch config to "true" pretty useless.

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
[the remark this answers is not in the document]
bffrost: @brianbolt I agree with you in general that we have a preference for lowecase true/false
However in looking at the historically loaded "data column order" experiment values (created by SEL), the "hide" and "condition" `stringValue`s are `TRUE` / `FALSE`.

I tried to match that pattern here, to keep the new data I insert consistent with preexisting data.

I noticed the same issue you've pointed out in ACASFormFields, and made this fix: https://github.com/mcneilco/acas/pull/1033/files#diff-21cee30a2860a441853c050a1d31946bdcf5945f65d07bcc5a1f5dbad1cf0420R1037

src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
```diff
@@ -0,0 +1,338 @@
...
+    select *, row_number() over (partition by experiment_id order by column_name) as column_order
+    from (select distinct experiment_id,
+                    ls_type as column_type,
+                    ls_kind as column_name,
+                    unit_kind as units,
+					tested_conc as concentration,
+                    tested_conc_unit as conc_units,
+					not public_data as hidden
+	-- we use the p_api_analysis_group_results view because it includes hidden data
+    from p_api_analysis_group_results paagr) a;
```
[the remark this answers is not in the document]
bffrost: I checked the performance of this on one of our largest staging servers, and both the API view and the base tables query were pretty performant (less than 2 minutes to run). The base tables query was more performant though, and it makes sense to filter for non-ignored data, so I've gone with the base tables.

bffrost:
I've tested this locally including with concentration and time. While on release/2022.4.x, I loaded these two experiment files:
Screen Shot 2022-11-15 at 1 11 17 PM

Then I switched onto this Roo branch, and the ACAS-411 ACAS branch, causing the migration to happen and the Endpoint Manager to be enabled. I then opened the protocol and confirmed all the endpoints from both experiments, including concentration and time information, were properly populated:
Screen Shot 2022-11-15 at 1 09 57 PM

brianbolt (approved):
LGTM

PR #395, opened 2022-11-10
ACAS-504: Improve BBChem bulk SDF upload performance

bffrost:
@brianbolt I'm still doing more testing and thinking more about this, but I think I have the "first draft". I'd like a review from you to see if there are major things I've missed, bugs you can spot, or better ideas you have based on what I've done so far.

## Description
- The basic observation here is that calling BBChem many times with small payloads is slow, and calling it few times with large payloads is fast. Thus the changes here try to take as many points where we call BBChem, and front-load them to doing calls in bulk.
- Read incoming SDF in large batches - during file / property scanning, property validation, and the dryrun/non-dryrun `registerSDF`.
- Call "standardize" i.e. "process" upfront and store the resulting CmpdRegMolecule as a transient property on the parent, pulling it out when needed rather than re-processing the molStructure string.
- Created a new method to call the "split" endpoint (which is part of getMolFragments) en masse. We have two usages of this: "checkForSalt" which returns a boolean, and "stripSalts" which actually does stuff with the split fragments. I've only optimized the "checkForSalt" portion, so we'll do a follow-up call to "split" as part of "stripSalts" for any records that do have multiple fragments.
- Switched the BBChem SDF writer to store a cache of molecules, then write everything out at once during the "close" or "getBufferString" method. I'm worried this may introduce bugs that make it hard to track down malformatted MOL blocks. I didn't do a great job with error handling.

## How Has This Been Tested?
Tested loading, purging, and re-loading a 999 record file on a remote LD/ACAS machine.
Compounds loaded successfully, with a ~20x speed improvement, and the summary files looked normal. Have not yet tested this locally in Indigo.

### DIFF
--- a/Dockerfile-multistage
+++ b/Dockerfile-multistage
@@ -7,7 +7,7 @@ ENV     CHEMISTRY_PACKAGE=${CHEMISTRY_PACKAGE}
 
 FROM 	dependencies as jchem
 COPY 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
-RUN   mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
+RUN 	--mount=type=cache,target=/root/.m2  mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
 
 FROM 	dependencies as bbchem
 
--- a/src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
@@ -592,7 +592,7 @@ public class BBChemStructureServiceImpl implements BBChemStructureService {
 	}
 
 	@Override
-	public List<String> getMolFragments(String molfile) throws CmpdRegMolFormatException {
+	public List<List<String>> getMolFragments(List<String> molfiles) throws CmpdRegMolFormatException {
 
 		String url = propertiesUtilService.getLDChemURL() + SPLIT_PATH;
 
@@ -602,15 +602,19 @@ public class BBChemStructureServiceImpl implements BBChemStructureService {
 
 		// Add the structures to the request
 		ArrayNode arrayNode = mapper.createArrayNode();
-		arrayNode.add(molfile);
+		for (String molfile : molfiles){
+			arrayNode.add(molfile);
+		}
 		requestData.replace("sdfs", arrayNode);
 
-		List<String> molStrings = new ArrayList<>();
+		List<List<String>> molStringsList = new ArrayList<List<String>>();
 		// Post to the service and parse the response
 		try {
 			String requestString = requestData.toString();
 			logger.debug("requestString: " + requestString);
+			logger.info("Making to bbchem split service");
 			HttpURLConnection connection = SimpleUtil.postRequest(url, requestString, logger);
+			logger.info("Got response from bbchem split service");
 			String postResponse = null;
 			if (connection.getResponseCode() != 200) {
 				logger.error("Error posting to split service: " + connection.getResponseMessage());
@@ -629,17 +633,19 @@ public class BBChemStructureServiceImpl implements BBChemStructureService {
 			JsonNode responseNode = responseMapper.readTree(postResponse);
 			JsonNode resultsNode = responseNode.get("results");
 			for (JsonNode resultNode : resultsNode) {
+				List<String> molStrings = new ArrayList<>();
 				JsonNode splitSDFS = resultNode.get("split_sdfs");
 				for (JsonNode sdfNode : splitSDFS) {
 					molStrings.add(sdfNode.asText());
 				}
+				molStringsList.add(molStrings);
 			}
 		} catch (Exception e) {
 			logger.error("Error posting to split service: " + e.getMessage());
 			throw new CmpdRegMolFormatException("Error posting to split service: " + e.getMessage());
 		}
 
-		return molStrings;
+		return molStringsList;
 	}
 
 	@Override
--- a/src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
@@ -510,7 +510,7 @@ public class BBChemStructureServiceImpl implements BBChemStructureService {
 		
 		// Post to the service and parse the response
 		String requestString = requestData.toString();
-		logger.info("Making to bbchem sdf export service");
+		logger.info("Making call to bbchem sdf export service");
 		logger.debug("requestString: " + requestString);
 		String postResponse = SimpleUtil.postRequestToExternalServer(url, requestString, logger);
 		logger.info("Got response from bbchem sdf export service");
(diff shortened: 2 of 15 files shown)


brianbolt:
I'll look at this PR in earnest tomorrow morning but as I see it, one thought I have for better testing coverage is to port 8010 from a bbchem server and just run tests as normal pointing at localhost:8010

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -382,23 +401,57 @@ public BulkLoadRegisterSDFResponseDTO registerSdf(BulkLoadRegisterSDFRequestDTO
 			CmpdRegSDFWriter registeredMolExporter = sdfWriterFactory.getCmpdRegSDFWriter(registeredSDFName);
 			HashMap<String, Integer> allAliasMaps = new HashMap<String, Integer>();
 
-			CmpdRegMolecule mol = null;
+			// CmpdRegMolecule mol = null;
```
brianbolt, line 404: Can we delete this instead

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -382,23 +401,57 @@ public BulkLoadRegisterSDFResponseDTO registerSdf(BulkLoadRegisterSDFRequestDTO
...
+			while (!done) {
+				// Read the next set of molecules from the SDF
+				Collection<CmpdRegMolecule> batchOfMols = molReader.readNextMols(batchSize);
+				if (batchOfMols.size() == 0){
+					done = true;
+					break;
+				}
+				// Process structures in bulk
+				HashMap<String, String> origMolStructures = new LinkedHashMap<String, String>();
+				HashMap<String, CmpdRegMolecule> origMols = new LinkedHashMap<String, CmpdRegMolecule>();
```
brianbolt, line 421: Wish I had looked up LinkedHashMap when refactoring for BBChem originally.  Nice to maintain order.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -801,7 +854,8 @@ public Parent validateParent(Parent parent, Collection<BulkLoadPropertyMappingDT
 			logger.warn(
 					"mol is empty and registerNoStructureCompoundsAsUniqueParents so not checking for dupe parents by structure but other dupe checking will be done");
 		} else {
-			dupeParentList = chemStructureService.checkDupeMol(parent.getMolStructure(), StructureType.PARENT);
+			dupeParentList = chemStructureService.searchMolStructures(parent.getCmpdRegMolecule(), StructureType.PARENT, ChemStructureService.SearchType.DUPLICATE_TAUTOMER, -1F, -1);
+			// dupeParentList = chemStructureService.checkDupeMol(parent.getMolStructure(), StructureType.PARENT);
```
brianbolt, line 864: Can we remove this comment

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -960,8 +1018,9 @@ public Parent validateParent(Parent parent, Collection<BulkLoadPropertyMappingDT
 	public Parent validateParentAgainstDryRunCompound(Parent parent, int numRecordsRead,
 			Collection<ValidationResponseDTO> validationResponse)
 			throws MissingPropertyException, DupeParentException, SaltedCompoundException, Exception {
-		int[] dupeDryRunCompoundsList = chemStructureService.checkDupeMol(parent.getMolStructure(),
-				StructureType.DRY_RUN);
+		// int[] dupeDryRunCompoundsList = chemStructureService.checkDupeMol(parent.getMolStructure(),
```
brianbolt, line 1027: Can we remove this comment?

src/main/java/com/labsynch/labseer/service/MetalotServiceImpl.java
```diff
@@ -313,33 +322,39 @@ public MetalotReturn processAndSave(Metalot metaLot, MetalotReturn mr, ArrayList
...
-				// multiple fragments
-				if (parent.getIsMixture() != null) {
-					if (!parent.getIsMixture()) {
+			} else {
+				Boolean hasMultipleFragments = parent.getMultipleFragments();
+				if (hasMultipleFragments == null){
+					hasMultipleFragments = chemService.checkForSalt(parent.getMolStructure());
+				}
+				if (hasMultipleFragments) {
+					// multiple fragments
```
brianbolt, line 331: Probably not the responsibility of this PR to fix this but just noting that it seems like the 2 inner if else blocks here can probably be skipped if we add them into line 330 properly.

brianbolt (approved):
Nice, I think pending getting acas client tests going for bbchem this LGTM.  I think it will be important to get this speed up in.

src/main/java/com/labsynch/labseer/service/MetalotServiceImpl.java
bffrost: Makes sense. I went ahead and did it, since I'm actively running the salts / fragments test.

bffrost:
I've done indigo and bbchem builds, and ensured all acasclient tests pass on both. Technically two of the bbchem tests fail, but they are for good reasons - one is due to a slight difference in SDF formatting from bbchem (adds a space after `<Parent Corp Name> `) and the other is a "large_sdf_with_error" whose error compound only errors in indigo.

I'd still like to ensure I can do a jchem build before merging this.

bffrost:
Got the JChem compile working. I can't do acasclient tests on JChem since I don't have a license, but calling this good enough and asking for a final re-review.

Dockerfile-multistage
```diff
@@ -7,7 +7,7 @@ ENV     CHEMISTRY_PACKAGE=${CHEMISTRY_PACKAGE}
 
 FROM 	dependencies as jchem
 COPY 	lib/jchem-16.4.25.0.jar /lib/jchem-16.4.25.0.jar
-RUN   mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
+RUN 	--mount=type=cache,target=/root/.m2  mvn install:install-file -Dfile=/lib/jchem-16.4.25.0.jar -DartifactId=jchem -DgroupId=com.chemaxon -Dversion=16.4.25.0 -Dpackaging=jar -DgeneratePom=true -DcreateChecksum=true
```
brianbolt, line 10: @philippcheung This will likely cause a conflict when you merge your branch but it should fix that issue where you had to comment out the cache to get the jchem builds working.

src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
```diff
@@ -510,7 +510,7 @@ public String getSDF(List<BBChemParentStructure> bbChemStructures) throws IOExce
 		
 		// Post to the service and parse the response
 		String requestString = requestData.toString();
-		logger.info("Making to bbchem sdf export service");
+		logger.info("Making call to bbchem sdf export service");
```
brianbolt, line 513: Thanks

brianbolt (approved):
LGTM!

src/main/java/com/labsynch/labseer/chemclasses/bbchem/BBChemStructureServiceImpl.java
```diff
@@ -602,15 +602,19 @@ public List<String> getMolFragments(String molfile) throws CmpdRegMolFormatExcep
...
 		requestData.replace("sdfs", arrayNode);
 
-		List<String> molStrings = new ArrayList<>();
+		List<List<String>> molStringsList = new ArrayList<List<String>>();
 		// Post to the service and parse the response
 		try {
 			String requestString = requestData.toString();
 			logger.debug("requestString: " + requestString);
+			logger.info("Making to bbchem split service");
 			HttpURLConnection connection = SimpleUtil.postRequest(url, requestString, logger);
```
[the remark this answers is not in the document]
bffrost: I double checked the python code, and it's doing a simple `for` loop through the "sdfs" inputs, so the order of inputs will be preserved in the output. I've tested this too.

PR #396, opened 2022-11-10
ACAS-504 Followup: Fix for writing SDF

bffrost:
## Description
I noticed in my testing after merging ACAS-504 that the "Download All Salts SDF" button in Salt Browser was writing the same salt out 3 times to the SDF. I traced it back to the fact that the code calls "close()" then calls "getBufferString" twice. This should be safe to do, so I fixed my modifications to CmpdRegSDFWriterBBChemImpl to clear out the molCache whenever "close" or "getBufferString" is called, so it's safe to call multiple times.

## How Has This Been Tested?
Tested locally that after this fix, the Salts SDF only had one entry per Salt.


brianbolt (approved):
Thanks, sorry I missed these things. 

LGTM

bffrost:
Very subtle bugs. Hopefully we got them all, but this one is definitely good to get QA on.

PR #401, opened 2022-11-17
ACAS-523: Skip throwing duplicate alias error if dupe is on the same parent

bffrost:
## Description
- Bug was that trying to load the same alias onto an existing parent would yield a duplicate alias error, rather than being a no-op
- Fix is to pass in the parent ID for the parent whose aliases are being validated, and skip reporting duplicates if the duplicate alias was found on the same parent

## How Has This Been Tested?
Ran acasclient tests locally with config `client.cmpdreg.metaLot.allowDuplicateParentAliases=false`
Confirmed that @dalejerikson 's `test_051_bulk_load_update_parent_alias` now passes

One other test fails with this config: `test_005_swap_parent_structures`, but that is expected because that case intentionally registers a duplicate alias to test an error thrown when an alias is ambiguous.

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -962,7 +962,7 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 								+ parent.getCorpName() + " db corp name: " + foundParent.getCorpName());
 		}
 		// Validate lot aliases locally and in the database
-		parentAliasService.validateParentAliases(parent.getParentAliases());
+		parentAliasService.validateParentAliases(parent.getId(), parent.getParentAliases());
 
 		return parent;
 	}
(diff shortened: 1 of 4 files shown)


iamanandkr (approved):
LGTM!

I am curious to know about how to handle the failing `test_005_swap_parent_structures` test - do we need ensure to not run the failing test when the config is set to not allow duplicates(based on does acasclient have access to the config value) or it's okay to let it fail?

bffrost:
@iamanandkr Yeah I think that is an open question. In GitHub Actions, the automated tests will always be run with our default configs, which do allow duplicates, so your test will pass. @brianbolt and I have tried to brainstorm before about how we might enable the automated CI tests to modify configs and run the tests with different sets of configs, but we haven't settled on a mechanism for doing that.

acasclient doesn't have programmatic access to the config values right now - we'd have to modify ACAS to expose some configs as an API endpoint to allow that.

dalejerikson (approved):
👍

PR #402, opened 2022-11-21
ACAS-489 Followup: Bug fixes for whitespace and for alias set comparisons

bffrost:
## Description
Fixed ACASClient tests (https://github.com/mcneilco/acasclient/pull/101) enough to uncover two more bugs:

### Semicolon-delimited aliases were not getting whitespace trimmed: 
```
>  <Parent Alias>
Fourth Alias; Fifth Alias; Sixth Alias; Seventh Alias;
```
parsed as (note leading whitespace):
```
["Fourth Alias"," Fifth Alias"," Sixth Alias"," Seventh Alias"]
```

### updateParentAliases was introducing duplicates
After fixing the above issue, noticed test was still failing. `compareParentAliases` function seems correct, but `updateParentAliases` function was not producing correct results, leading to duplicate aliases getting created.
The logical flow inside that function was confusing  enough that I rewrote it to be able to result in the correct behavior.

## How Has This Been Tested?
Ran acasclient tests locally and confirmed this fixes the test failures on test_051 mentioned here https://github.com/mcneilco/acasclient/pull/101

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
@@ -2392,31 +2393,30 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 		// Note: This is not the most efficient / concise code to do this process
 		// however due to readability and the assumingly small number of aliases 
 		// parents will have this is fine 
-
-		Set<String> parentAliasStrings = new HashSet<>();
+		HashMap<String, ParentAlias> newAliases = new HashMap<String, ParentAlias>();
 		for(ParentAlias alias : parent.getParentAliases()){
 			if (! alias.isIgnored()) {
-				parentAliasStrings.add(alias.getAliasName());
+				newAliases.put(alias.getAliasName(), alias);
 			}
 		}
-		Set<String> foundParentAliasStrings = new HashSet<>(); 
+		HashMap<String, ParentAlias> existingParentAliases = new HashMap<String, ParentAlias>();
 		for(ParentAlias alias : foundParent.getParentAliases()){
 			if (! alias.isIgnored()) {
-				foundParentAliasStrings.add(alias.getAliasName());
+				existingParentAliases.put(alias.getAliasName(), alias);
 			}
 		}
-
-		Set<ParentAlias> unionParentAliases = parent.getParentAliases();
+		// Create a set to hold the union of all unique aliases
+		// Start with the existing aliases
+		Set<ParentAlias> unionParentAliases = foundParent.getParentAliases();
+		// Find any new aliases not already in the old set, and add them
+		Set<String> newAliasStrings = newAliases.keySet().stream().filter(e -> 
+			!existingParentAliases.keySet().contains(e)).collect(Collectors.toSet());
+		for(String newAliasString : newAliasStrings) {
+			unionParentAliases.add(newAliases.get(newAliasString));
+		}
 		// Parent Would Be Found Parent w/ Appropriate Updates
 		parent = foundParent;
 		// Update Parent Object to Have "Union" of All Aliases 
-		for(ParentAlias oldAlias : parent.getParentAliases()) {
-			// If oldAlias Not In UnionParentAliases (Use Previously Set<str> to Do Compare)
-				// Add to Aliases Union List 
-			if(! oldAlias.isIgnored() && !parentAliasStrings.contains(oldAlias.getAliasName())) {
-				unionParentAliases.add(oldAlias);
-			}
-		}
 		parent.setParentAliases(unionParentAliases);
 		// If Alias List Updated Then Updated Modified By and Modified Date 
 		parent.setModifiedDate(new Date());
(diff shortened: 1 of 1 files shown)


brianbolt (approved):
LGTM

dalejerikson:
@bffrost I know this was already merged while I'm away but I was curious why the previous implementation of updateParentAliases was buggy and what part of the code allowed duplicates to slip through? I just skimmed it but theoretically the original implementation used a Set of strs and seemed correct?

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2391,31 +2393,30 @@ private Parent updateParentAliases(Parent parent, Parent foundParent, String che
 		// Note: This is not the most efficient / concise code to do this process
 		// however due to readability and the assumingly small number of aliases 
 		// parents will have this is fine 
-
-		Set<String> parentAliasStrings = new HashSet<>();
```
bffrost, line 2395: This is a set of the new incoming alias strings

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2391,31 +2393,30 @@ private Parent updateParentAliases(Parent parent, Parent foundParent, String che
...
+		HashMap<String, ParentAlias> newAliases = new HashMap<String, ParentAlias>();
 		for(ParentAlias alias : parent.getParentAliases()){
 			if (! alias.isIgnored()) {
-				parentAliasStrings.add(alias.getAliasName());
+				newAliases.put(alias.getAliasName(), alias);
 			}
 		}
-		Set<String> foundParentAliasStrings = new HashSet<>();
```
bffrost, line 2401: This is a set of the old existing alias strings

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2391,31 +2393,30 @@ private Parent updateParentAliases(Parent parent, Parent foundParent, String che
...
-		Set<String> foundParentAliasStrings = new HashSet<>(); 
+		HashMap<String, ParentAlias> existingParentAliases = new HashMap<String, ParentAlias>();
 		for(ParentAlias alias : foundParent.getParentAliases()){
 			if (! alias.isIgnored()) {
-				foundParentAliasStrings.add(alias.getAliasName());
+				existingParentAliases.put(alias.getAliasName(), alias);
 			}
 		}
-
-		Set<ParentAlias> unionParentAliases = parent.getParentAliases();
```
bffrost, line 2408: We make a set of the new *ParentAlias*es. These all lack the `id` property, i.e. they're all new Java objects that have not been persisted to the DB.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2391,31 +2393,30 @@ private Parent updateParentAliases(Parent parent, Parent foundParent, String che
...
+		// Start with the existing aliases
+		Set<ParentAlias> unionParentAliases = foundParent.getParentAliases();
+		// Find any new aliases not already in the old set, and add them
+		Set<String> newAliasStrings = newAliases.keySet().stream().filter(e -> 
+			!existingParentAliases.keySet().contains(e)).collect(Collectors.toSet());
+		for(String newAliasString : newAliasStrings) {
+			unionParentAliases.add(newAliases.get(newAliasString));
+		}
 		// Parent Would Be Found Parent w/ Appropriate Updates
 		parent = foundParent;
```
bffrost, line 2418: Then we swap the meaning of "parent"

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2391,31 +2393,30 @@ private Parent updateParentAliases(Parent parent, Parent foundParent, String che
...
 		// Update Parent Object to Have "Union" of All Aliases 
-		for(ParentAlias oldAlias : parent.getParentAliases()) {
```
bffrost, line 2412: Then we loop through the existing aliases, and only add ones whose string is not found in the new aliases.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -2391,31 +2393,30 @@ private Parent updateParentAliases(Parent parent, Parent foundParent, String che
...
-			// If oldAlias Not In UnionParentAliases (Use Previously Set<str> to Do Compare)
-				// Add to Aliases Union List 
-			if(! oldAlias.isIgnored() && !parentAliasStrings.contains(oldAlias.getAliasName())) {
-				unionParentAliases.add(oldAlias);
-			}
-		}
 		parent.setParentAliases(unionParentAliases);
```
bffrost, line 2420: @dalejerikson 
The result is that we'll have:
- old non-overlap ParentAliases
- new non-overlap ParentAliases
- overlapping ones represented by the new Java object (i.e. with no `id`)

The old overlapping aliases are not included in this set, but they still exist in the database. They don't get ignored or deleted just by saving the `parent` with a set of `parentAliases` that doesn't include them.

The result is that any overlapping aliases get duplicated. We save the new copies of them as new rows in `parent_alias` and the old copies of them still exist in `parent_alias`.

bffrost (commented):
@dalejerikson Very valid question, and this was a subtle bug that I missed in code review, and we only caught through testing. I've added an explanation.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
bffrost: The main difference between my implementation and yours is that I preferentially use the *old* ParentAlias objects and only add "new non-overlap ParentAliases"

PR #404, opened 2022-11-27
ACAS-526: Delete standardization dry run compound when deleting a parent compound

brianbolt:
## Description
 - Delete standardization dry run compound when deleting a parent compound

## Related Issue
ACAS-526

## How Has This Been Tested?
- Test locally to reproduce
  - Ran acas client test with breakpoint to register an SDF file (then stopped the run so teardown was run)
  - Executed standardization dry run
  - Tried to delete one of the lots (single lot on parent so the parent is also attempted to be deleted)
    - Result was failure to delete lot
 - Test fix
   -  updated the code, ran the same test and now the parent is deleted successfully (including from standardization dry run table)


dalejerikson (approved):
👍

bffrost (approved):
LGTM

bffrost:
Good catch

PR #405, opened 2022-12-01
ACAS-413: DB migration for endpoint management (2nd attempt)

bffrost:
## Description
- Earlier PR #394 was merged but caused issues and was reverted
- Primary issue was that it failed on brand new / empty systems due to some missing sequences that are created when ACAS starts up and runs setup routes
- Fixed this issue by wrapping the entire script in a "DO" block and thus turning it into an anonymous pl/pgsql function. This allows us to return early if no experiment data exists.
- While locally testing upgrade from 2022.4.x to this branch, found a few more breaking issues and fixed them. They were missing ddict types and kinds, a copy/paste typo where column name was put in instead of units, and some followthrough on switching to pl/pgsql (replacing lone "select" statements with "perform")

I'm sorry this is displaying as one big change, but you can look through the individual commits to get a sense of the individual changes that separate this from the earlier PR #394 

## How Has This Been Tested?
Tested two scenarios locally:
1. Start up local ACAS on latest 2022.4.x, load compounds and an experiment, shut down, switch to 2023.1.x + this Roo branch, and start up. Ensured migration ran successfully, and checked Protocol and Experiment browsers and ensured all the expected content was in the endpoints table in both.
2. Started up completely fresh system (`docker-compose down -v && docker-compose up`) on 2023.1.x + this roo branch, and ensured it started up smoothly and acasclient tests passed.

### DIFF
--- a/src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
+++ b/src/main/resources/db/migration/postgres/V2.4.1.0__data_column_order_states.sql
@@ -41,6 +41,24 @@ insert into ls_transaction (id, comments, recorded_date, version, recorded_by)
 -- Part 1: Fill in 'data column' DDict Values based on saved data
 -- Create a reusable function to insert missing ddict values
 create or replace function add_data_column_ddict_value(val varchar, ls_kind varchar) returns bigint as $$
+	--create ddict type if not exists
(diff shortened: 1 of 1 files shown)


brianbolt (approved):
LGTM

PR #407, opened 2022-12-07
ACAS-413: look up ddict id sequence rather than assuming it

bffrost:
## Description
- Some older servers seem to have a sequence named:
`labelseq_DDICT_id_codeName_document_datadictionary` instead of
`labelseq_1_ddict_id_codename_document_datadictionary`
- This change actually looks up the db_sequence name in the label_sequence table rather than assuming we know the name.

## How Has This Been Tested?
- Tested the SQL command runs on an older and a newer ACAS server


brianbolt (approved):
LGTM

PR #410, opened 2022-12-16
ACAS-567: Trim empty strings in incoming aliases

bffrost:
## Description
- SDF with some Parent Aliases populated, others blank, and others missing that property entirely were failing to register when `allowDuplicateParentAliases` was set to false.
- Root issue seems to be we were passing through empty-string aliases and trying to create them
- Fixed it by filtering out empty strings and nulls after the string is split and trimmed
- Also removed some misleading logging

## How Has This Been Tested?
- Reproduced the issue locally and tested with several combinations of files, and with combinations of:
- registering new compounds
- registering new lots of existing compounds with & without aliases
- added acasclient test

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -17,7 +17,9 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -1793,9 +1783,15 @@ public Lot addLotAlias(Lot lot, String lsType, String lsKind,
 			String lookUpProperty, String lookUpString) {
 
 		if (lookUpString != null) {
-			// found LiveDesign Corp Name
-			logger.info("Found one or more lot alias: " + lookUpProperty + "  " + lookUpString);
-			String[] aliases = lookUpString.split(";");
+			// Split on semicolon and trim
```
brianbolt, line 1786: nit* - Because we are doing this twice, can we wrap it in a helper function?

brianbolt (approved):
Nit about duplicated code but LGTM otherwise.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
bffrost: Yep, done!

PR #411, opened 2023-01-04
ACAS-413: Add a new migration to catch any "missed" ddict values.

bffrost:
## Description
- Testing showed that some "Dose Response" columns (Slope, Min, Max) present in SEL files were being saved in Experiment "data column order" states by SEL, but the necessary DDict values to display their column names in the Protocol Endpoint Manager GUI were not getting created
- Simplest fix is to add a new migration that looks for any missing ddict values and adds them

## How Has This Been Tested?
Tested locally by:
- Starting up ACAS on release/2022.4.x
- Loading a Dose Response Experiment
- Switching ACAS to release/2023.1.x and starting up, which applies the previous migration V2.4.1.0
- Logging in, navigating to Protocol Browser and confirming there are endpoints whose names are not displaying
- Switch Roo image to locally-built one with these changes, start up, which applies the new migration V2.4.1.1
- Login again and check the Protocol UI, and confirm all endpoints are now showing their column names properly.

### DIFF
--- a/src/main/resources/db/migration/postgres/V2.4.1.1__missing_column_ddict_values.sql
+++ b/src/main/resources/db/migration/postgres/V2.4.1.1__missing_column_ddict_values.sql
@@ -1,0 +1,116 @@
+do $script$
+begin
+-- As part of ACAS-413, followup from previous migration
+-- Exit early if there is no expt data as this indicates a fresh system
+if not exists (select * from experiment)
+then
+	return;
+end if;
+
+-- Fill in missing 'data column' DDict Values based on saved data
+-- Create a reusable function to insert missing ddict values
+create or replace function add_data_column_ddict_value(val varchar, ls_kind varchar) returns bigint as $$
+	--create ddict type if not exists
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/postgres/V2.4.1.1__missing_column_ddict_values.sql
```diff
@@ -0,0 +1,116 @@
...
+					and dv.ls_kind = ls_kind 
+					and dv.ignored = false 
+					and dv.short_name = val
+			)
+		returning id
+	$$ language sql;
+
+--Add missing column name ddicts
+perform add_data_column_ddict_value(code_value, 'column name'::varchar) from
+(SELECT DISTINCT ev.code_value
```
bffrost, line 54: The key new query is this simple one here, which looks for "column name" experiment values within the "data column order" state, for which a corresponding ddict_value does not exist

brianbolt (approved):
LGTM

PR #415, opened 2023-01-27
ACAS-601: Fix for multiple calls to split service for already registered parents

brianbolt:
## Description
 - Maintain transient multipleFragments property when finding an already register compound
 - Don't do multiple fragment errors when finding an already registered parent

## Related Issue
ACAS-601

## How Has This Been Tested?
Not tested yet

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -980,7 +980,7 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 		}
 
 		// Only run mixture check if we have not identified an already existing parent
-		if(parent.getId() != null) {
+		if(parent.getId() == null) {
 			// check for multiple fragments, and if isMixture or not
 			Boolean hasMultipleFragments = parent.getMultipleFragments();
 			if (hasMultipleFragments == null){
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -868,7 +868,7 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 		if (dupeParentList.length > 0) {
 			// If we find a duplicate parent, we want to maintain the multiple fragments boolean we already have to save time since we have already checked for this
 			// and it is a transient property which wil be null on the found parent we may be overwriting with
-			boolean multipleFragments = parent.getMultipleFragments();
+			Boolean multipleFragments = parent.getMultipleFragments();
 			searchResultLoop: for (int foundParentCdId : dupeParentList) {
 				List<Parent> foundParents = Parent.findParentsByCdId(foundParentCdId).getResultList();
 
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -866,6 +866,9 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 			dupeParentList = chemStructureService.searchMolStructures(parent.getCmpdRegMolecule(), StructureType.PARENT, ChemStructureService.SearchType.DUPLICATE_TAUTOMER, -1F, -1);
 		}
 		if (dupeParentList.length > 0) {
+			// If we find a duplicate parent, we want to maintain the multiple fragments boolean we already have to save time since we have already checked for this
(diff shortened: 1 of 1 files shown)


brianbolt:
I have not tested this or even compiled it yet.  I saw this while running some tests to check why K8s ACAS is slow.  This is not why K8s is slow but saw this extra slowdown on loading the same file multiple times.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -972,25 +978,29 @@ public Parent validateParent(Parent parent, String chemist, Collection<BulkLoadP
...
-		if (hasMultipleFragments == null){
-			hasMultipleFragments = chemStructureService.checkForSalt(parent.getMolStructure());
-		}
-		if (hasMultipleFragments) {
-			// multiple fragments
-			if (parent.getIsMixture() != null) {
-				if (!parent.getIsMixture()) {
+
+		// Only run mixture check if we have not identified an already existing parent
+		if(parent.getId() != null) {
```
bffrost, line 983: I'm confused by your comment plus this line:
"Only run mixture check if we have not identified an already existing parent"
An existing parent would have id != null
So you go into the `if` clause only for pre-existing parents, and then run the multiple fragment check?

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -866,6 +866,9 @@ public Parent validateParent(Parent parent, String chemist, Collection<BulkLoadP
 			dupeParentList = chemStructureService.searchMolStructures(parent.getCmpdRegMolecule(), StructureType.PARENT, ChemStructureService.SearchType.DUPLICATE_TAUTOMER, -1F, -1);
 		}
 		if (dupeParentList.length > 0) {
+			// If we find a duplicate parent, we want to maintain the multiple fragments boolean we already have to save time since we have already checked for this
+			// and it is a transient property which wil be null on the found parent we may be overwriting with
+			boolean multipleFragments = parent.getMultipleFragments();
```
bffrost, line 871: I think technically this should be a `Boolean` (allowing for null values) even though we don't expect this to be null at this point.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -894,6 +897,9 @@ public Parent validateParent(Parent parent, String chemist, Collection<BulkLoadP
 							// No Changes or Modifications Needed; Can Break Loop
 							// Continue 
 							parent = foundParent;
+							// Add the multiple fragments boolean back to the parent
+							// from the transient property we calculated earlier
+							parent.setMultipleFragments(multipleFragments);
```
bffrost, line 902: This part makes sense to me - if it's the same structure, we can copy in this property without hitting BBChem again.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
```diff
@@ -972,25 +978,29 @@ public Parent validateParent(Parent parent, String chemist, Collection<BulkLoadP
 				}
 			}
 		}
-		// check for multiple fragments, and if isMixture or not
```
bffrost, line 975: I'm confused by all the changes down in this section. As far as I can tell, you wrapped the previous logic in an `if(parent.getId() != null` block. So we're only checking pre-existing parents. Maybe if you reversed that logic it would make more sense - so we'd only be checking new parents.

bffrost (changes requested):

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
brianbolt: Thanks. That was wrong. I fixed it.

src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
brianbolt: Thanks, should have noticed the underlying type. Fixed it.

bffrost (approved):
LGTM

bffrost:
@brianbolt I tested this out and it seems to work well.
- Compiles fine
- Confirmed on 2nd load of a large file that it eliminates the one-by-one calls to split service
- Ran acasclient test_047_load_sdf_with_salts which does involve some salt splitting
- Also loaded the file from test_047_load_sdf_with_salts manually, mapping in Parent Is Mixture: true, and loading that file multiple times, including after manually deleting some of the multi-fragment lots. AFAICT it seemed to load everything properly.

PR #416, opened 2023-01-27
ACAS-600: Reparent lot behavior with maxAutoLotNumber

bffrost:
Catch case when there are no lots below maxAutoLotNumber and return 0…… as the max Lot Number. This prevents upstream NullPointerExceptions.

## Description
- The scenario for this bug is when a user is trying to reparent a lot (i.e. lot 1) of a Parent A onto another Parent B which only has a huge lot number (i.e. lot 4000). When the `maxAutoLotNumber` config is set (i.e. to 1000) the expected behavior is that Parent A Lot 1 will become Parent B Lot 1.
- On ACAS 2022.1.0 the actual behavior is that Parent A Lot 1 will be made into Parent B Lot 4001
- On ACAS 2022.4.3 the actual behavior is that the reparent check fails in the UI, and on the Roo layer there is a NullPointerException.
- This PR contains the fix for the NullPointerException. After that fix, we see the expected behavior.

## How Has This Been Tested?
- Tested in local dev environment.

### DIFF
--- a/src/main/java/com/labsynch/labseer/domain/Lot.java
+++ b/src/main/java/com/labsynch/labseer/domain/Lot.java
@@ -405,7 +405,11 @@ public class Lot {
         }
 		criteria.where(criteriaBuilder.and(predicateList.toArray( new Predicate[0])));
         TypedQuery<Integer> q = em.createQuery(criteria);
-        return q.getSingleResult();
+        Integer result = q.getSingleResult();
+        if(result == null) {
+            result = 0;
+        }
+        return result;
     }
 
     @SuppressWarnings("unchecked")


brianbolt (approved):
LGTM. Might be worth adding some acasclient tests.  I'm pretty sure it was some time after I "fixed/broke" this code that we set ACAS by default to have `client.cmpdreg.metaLot.maxAutoLotNumber=1000`

bffrost:
@brianbolt Totally agree. I've got acasclient tests and just reproduced this issue using them. Confirmed with this fix that they fully pass. I've opened up https://github.com/mcneilco/acasclient/pull/126
I'm going to confirm that test fails, merge this Roo change, then rerun the acasclient test and it should go green.

PR #417, opened 2023-01-27
ACAS-603: Fix standardization status on fresh system

bffrost:
### To reproduce
- Start up a fresh instance of ACAS
- Bulk load some compounds
- Open up Bulk loader or CmpdReg
### Expectation
ACAS does not block you from registering compounds and requires a re-standardization
### Actual outcome
ACAS blocks you from registering compounds and requires re-standardization

## Description of fix
- Bug was caused by the fact our auto-created initial standardizer history had no settings / settings hash
- Getting the applied settings and saving them in that initial history entry solves this issue

## How Has This Been Tested?
- Reproduced issue in k8s and in local dev (attached to remote bbchem)
- Confirmed on local dev (attached to remote bbchem) that this is fixed after the code changes


brianbolt (approved):
LGTM. This is actually what I thought it might be too. Wonder if this also fixes the issue I saw sporadically with the 2022.4.3 standardization execution not "sticking".

bffrost:
@brianbolt I was wondering the same thing about the not "sticking" issue.

PR #418, opened 2023-01-28
ACAS-602 delete standardization dry run compounds when purging a bulk load file

brianbolt:
## Description
 - Check for and delete standardization dry run compounds when deleting a bulk load file

## Related Issue
ACAS-602

## How Has This Been Tested?
Ran standardization dry run
Purged a bulk load file and verified it was successful and that the rows associated with the parents were also deleted in standardization_dry_run table.


bffrost (approved):
LGTM

PR #419, opened 2023-01-30
ACAS-596: Health endpoint to check database speed.  Better logging of bbchem process and fingerprint requests

brianbolt:
## Description
 - Health endpoint queries all value types, which should be uniform on all systems of the same version. The idea is to have an endpoint tests the speed of querying the database which can be used to compare systems.

Curl example:

```
curl  "http://localhost:8080/acas/api/v1/health/checkDatabaseSpeed?testsToRun=100000&logData=False"
```

Output:
```
{"averageMsTime": 0.73629, "totalTime": 73629, "testsToRun": 100000, "logData": false}
```

## Related Issue
ACAS-596

## How Has This Been Tested?
Ran curl and verified results with up to 1 million tests with log data False and True.

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiHealthController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiHealthController.java
@@ -1,0 +1,68 @@
+package com.labsynch.labseer.api;
+
+import java.util.Collection;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


bffrost (approved):
LGTM, improved logging makes sense and database speed health check makes sense

PR #420, opened 2023-02-01
ACAS-596: Use parallelStream when running dupeCheckStandardizationStructuresBatch

brianbolt:
## Description
 - Run parallelStream when checking for duplicate structures
 
I saw the following improvements when running in K8s with 1CPU request and 3CPU limit

Before
```
populating standardization dry run table 100.0% complete (5000 of 5000) average speed (rows/min):4113.3643206778825
checking for standardization duplicates 100.0% complete (5000/5000) average speed (rows/min):6638.0549963498
```

After
```
populating standardization dry run table 80.0% complete (4000 of 5000) average speed (rows/min):4101.372250798913
checking for standardization duplicates 100.0% complete (5000/5000) average speed (rows/min):18230.755209915547
```

## Related Issue
ACAS-596

## How Has This Been Tested?
Ran through standardization with 5000 compound set on K8s.  There were no duplicates on the system so that branch of the code was not tested with this change.

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
@@ -401,7 +401,7 @@ public class StandardizationServiceImpl implements StandardizationService, Appli
 
 	@Override
 	@Transactional
-	public int dupeCheckStandardizationStructures() throws CmpdRegMolFormatException {
+	public int dupeCheckStandardizationStructures() {
 		List<Long> dryRunIds = StandardizationDryRunCompound.findAllIds().getResultList();
 
 		Long startTime = new Date().getTime();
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -401,7 +401,7 @@ private String getParentAlias(Parent parent) {
 
 	@Override
 	@Transactional
-	public int dupeCheckStandardizationStructures() throws CmpdRegMolFormatException {
```
bffrost, line 404: This compiles properly, right? So we don't need to change the abstract interface at StandardizationService?

I recall having to tweak both places, but maybe it is OK with an implementation that just doesn't throw some of the exception types.

bffrost (approved):
The parallelStream stuff looks great. Had a question about the change to thrown exceptions but as long as it compiles this LGTM

PR #421, opened 2023-02-08
Salt Browser Bug: Charged and Uncharged Salts Being Treated as Duplicates 

dalejerikson:
**Bug:** CmpdReg Salt Module: Different forms of same salt (e.g. charged & neutral) are being treated as duplicates

Steps to Reproduce:

- Log in to ACAS as CmpdReg admin >> Open CmpdReg Salts module
- Register a salt in it's neutral form (e.g. Dihydrogen Sulphate - H2SO4)
- Now register a charged form of the same salt (i.e. Sulphate ion - O4S-2)
- Provide different salt name & abbreviation for both entries

Previous Bug Results:

- The neutral form & charged form of same salt is getting treated as duplicates of each other
- CReg is generating '0' Compound ID for the later entry

Fixed Results:

- Both entries should get registered as different entries
- Both entries should receive unique Compound ID

Extra Information: 

- Added `saveStructureNoDupeCheck(salt)` in `saltStructureService` 

Logic and Reasoning for Solution: 

- When first implementing the Salt Browser, it was decided that we would not be checking for duplicate structures and it would the responsibility of the data administrators to retain the integrity of the data in this area. This work uncovered a prior section of code salts would be run through when saved: 

```
boolean checkForDupes = true;
int cdId = chemStructureService.saveStructure(salt.getMolStructure(), StructureType.SALT, checkForDupes);
```

- Different chemistry engines slightly vary the way in which duplicates are checked and since the original design of the Salt Browser purposefully did not involve duplicate checking, I have added `saveStructureNoDupeCheck(salt)` to bypass this problem.

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiSaltController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiSaltController.java
@@ -283,7 +283,7 @@ public class ApiSaltController {
 		}
 		if (validSalt & !dryrun) {
 			try {
-				salt = saltStructureService.saveStructureNoDupeCheck(salt);
+				salt = saltStructureService.saveStructure(salt, false);
 			} catch (CmpdRegMolFormatException e) {
 				logger.error("Error saving salt: " + e.getMessage());
 				validSalt = false;
--- a/src/main/java/com/labsynch/labseer/api/ApiSaltController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiSaltController.java
@@ -283,7 +283,7 @@ public class ApiSaltController {
 		}
 		if (validSalt & !dryrun) {
 			try {
-				salt = saltStructureService.saveStructure(salt);
+				salt = saltStructureService.saveStructureNoDupeCheck(salt);
 			} catch (CmpdRegMolFormatException e) {
 				logger.error("Error saving salt: " + e.getMessage());
 				validSalt = false;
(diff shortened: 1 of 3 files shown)


src/main/java/com/labsynch/labseer/api/ApiSaltController.java
```diff
@@ -283,7 +283,7 @@ public ResponseEntity<String> createFromJson(@RequestBody String json, @RequestP
 		}
 		if (validSalt & !dryrun) {
 			try {
-				salt = saltStructureService.saveStructure(salt);
+				salt = saltStructureService.saveStructureNoDupeCheck(salt);
```
brianbolt, line 286: I think it would be preferrable to reuse the logic in saveStructure but just add a parameter "checkDuplicates" to the method.  This would involved making the other calls to `saveStructure` pass `true` to checkDuplicates but this is reasonable to do as I don't think there are many instances.

brianbolt (changes requested):

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
dalejerikson: Yea I was debating on it but was on the fence. Definitely can do that.

bffrost:
Looks good aside from Brian's comment about consolidating the saveStructure & saveStructureNoDupeCheck methods.

src/main/java/com/labsynch/labseer/api/ApiSaltController.java
dalejerikson: For some reason thought there were more instances of the saveStructure utilized by SaltStructureService. My bad.

bffrost (approved):
LGTM! No more dupe checking of salts!

PR #423, opened 2023-02-09
k8s: CmpdReg Salts module is not calculating "charges" from salt's structure while registering salts 

dalejerikson:
**Steps to Reproduce:**

1. Be using an ACAS on k8s instance
2. Log in to ACAS as CmpdReg Admin
3. Open "CmpdReg Salts" module >> Click 'Create New Salt'
4. Draw a salt structure with '+' or '-' charges (e.g. sulphate) >> add Abbreviation & Name
5. Click on "Next"
6. Observe the Salt Charge & Salt Molecular Weight calculated on next page

**Expected Results:**

- The molecular weight is calculated for "charged" form of the salt & not the neutral one
- The "Salt Charge" is generated as per the drawn structure

**Previous Bug Results:**

- Molecular weight is generating for neutral form
- Salt charge is not computed (showing '0')

### DIFF
--- a/src/main/java/com/labsynch/labseer/chemclasses/CmpdRegMolecule.java
+++ b/src/main/java/com/labsynch/labseer/chemclasses/CmpdRegMolecule.java
@@ -19,13 +19,13 @@ public interface CmpdRegMolecule {
 
 	public String getMolStructure() throws CmpdRegMolFormatException;
 
-	public String getFormula();
+	public String getFormula(boolean stopNeutralization);
 
 	public Double getExactMass();
 
-	public Double getMass();
+	public Double getMass(boolean stopNeutralization);
 
-	public int getTotalCharge();
+	public int getTotalCharge(boolean stopNeutralization);
 
 	public String getSmiles();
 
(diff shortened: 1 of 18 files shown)


bffrost:
@dalejerikson Just an idea - could we add a new "skipNeutralization" flag to CmpdRegMolecule itself, and default it to false?
Then have Salts set that to true? It isn't very different from your approach, it just might make for fewer code changes to store that flag on the molecule itself, then check it down within the `calculateCharge`, `calculateFormula`, `calculateWeight` functions. I think you won't be able to avoid adding it to `getProcessedStructure` since that takes a String molfile. But it might reduce the number of required changes.

dalejerikson:
Any additional comments/thoughts @brianbolt ?

dalejerikson:
> @dalejerikson Just an idea - could we add a new "skipNeutralization" flag to CmpdRegMolecule itself, and default it to false? Then have Salts set that to true? It isn't very different from your approach, it just might make for fewer code changes to store that flag on the molecule itself, then check it down within the `calculateCharge`, `calculateFormula`, `calculateWeight` functions. I think you won't be able to avoid adding it to `getProcessedStructure` since that takes a String molfile. But it might reduce the number of required changes.

I'll check it out and see what that would like!

brianbolt:
I am on the same page as you. I think there could be a more elegant solution but I don't see an obvious way to do that without more work.  My one nit is I think `stopNeutralization` could be more clear as `forceNeutralizationFalse` but other than that I think this LGTM.

brianbolt:
I turned this from a draft to "Ready for Review"

brianbolt (approved):
LGTM

dalejerikson:
@brianbolt I think you actually beat me to this by a minute lol

PR #424, opened 2023-02-14
Replacing "false" with false when setting NEUTRALIZE config 

dalejerikson:
In some instances, the standardizer actions will not recognize "false" as valid and will throw an error that 

`"NEUTRALIZE must have be a boolean value: false"`

This ultimately can lead to an error where `Request method 'POST' not supported` and appears to the user as a misleading `405` error. 

Setting this value directly as a bool value `false` will prevent this problem from happening. 

Note: This came up from this code introduced here: https://github.com/mcneilco/acas-roo-server/pull/423


bffrost (approved):
LGTM

PR #433, opened 2023-04-01
ACAS-644: Update label sequence to include unique constraint across thing_type_and_kind, label_type_and_kind, label_prefix

brianbolt:
## Description

On start the `acas` container sends a series of requests to `roo` `/setup/:entity` routes https://github.com/mcneilco/acas-roo-server/blob/cdfadece229047c166b02e52f9ce88b509938c0e/src/main/java/com/labsynch/labseer/api/ApiSetupController.java#L291 
ACAS does this to load default label sequences, value types...etc. into the DB.  When ACAS runs with multiple replicas it can end up running the e.g. `/setup/labelsequences` route at the exact same time and because hibernate and the database are unaware of our business logic constraint on label sequences found here https://github.com/mcneilco/acas-roo-server/blob/e7bde99239fbb0e628962c04ac41bdd09542e93b/src/main/java/com/labsynch/labseer/service/LabelSequenceServiceImpl.java#L37 it will load multiple of the same label sequences which breaks login and other things

## Details on fix
- This fix just implements a partial unique index for both the `label_sequence` and `ddict_value` tables to prevent duplicates from being added. 
- I chose not to fix the fact that each replica of ACAS sends a request to setup defaults because this is a bigger change to the infrastructure and to the way ACAS has worked for a long time.
- Only `/setup/labelsequences` and `/setup/codetables` could actually cause issues as the other routes have database constraints which would prevent this from happening.

## Related Issue
ACAS-644

## How Has This Been Tested?
**Script:**
I wrote a script which consistently reproduced the issue before this fix.  It does this:
- Truncates the label sequences table
- Deletes both pods at about the same time and then waits for them to come back up fully
- Checks for the issue by running the query above
- Repeats until the issue is present.
I do not see the issue after the fix.

**Manual symptom check**
Verified I could login on a fresh system with this fix. 

**Other checks**
 - With AUTH sequence in place in label sequences I tried to insert a duplicate manually with this:
 ```
 INSERT INTO label_sequence (id, digits, group_digits, ignored, label_prefix, label_separator, label_type_and_kind, starting_number, modified_date, thing_type_and_kind, version, db_sequence)
    VALUES (nextval('hibernate_sequence'), 8, FALSE, FALSE, 'AUTH', '-', 'id_codeName', 1, '2019-01-01 00:00:00', 'author_author', 0, 'labelseq_1890_AUTH_id_codeName_author_author');
```

This failed (as expected with):
```
ERROR:  duplicate key value violates unique constraint "label_sequence_thing_label_prefix_key"
DETAIL:  Key (thing_type_and_kind, label_type_and_kind, label_prefix)=(author_author, id_codeName, AUTH) already exists.
```

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/LabelSequenceServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/LabelSequenceServiceImpl.java
@@ -36,7 +36,7 @@ public class LabelSequenceServiceImpl implements LabelSequenceService {
 						.findLabelSequencesByThingTypeAndKindEqualsAndLabelTypeAndKindEquals(
 								labelSequence.getThingTypeAndKind(), labelSequence.getLabelTypeAndKind())
 						.getSingleResult();
-				logger.info("Found existing LabelSequence: '" + savedLabelSequence.getThingTypeAndKind() + "'' "
+				logger.info("Found existing LabelSequence: '" + savedLabelSequence.getThingTypeAndKind() + "' '"
 						+ savedLabelSequence.getLabelTypeAndKind() + "'");
 			} catch (NoResultException e) {
 				logger.info("No existing LabelSequence found, creating new one");

### CODE REVIEW

File: src/main/java/com/labsynch/labseer/service/LabelSequenceServiceImpl.java [hunk 1]
```diff
@@ -30,11 +30,16 @@ public Collection<LabelSequence> saveLabelSequenceArray(
 		for (LabelSequence labelSequence : labelSequences) {
 			LabelSequence savedLabelSequence;
 			try {
+				logger.info("Looking for existing LabelSequence: '" + labelSequence.getThingTypeAndKind() + "' '"
+						+ labelSequence.getLabelTypeAndKind() + "'");
 				savedLabelSequence = LabelSequence
 						.findLabelSequencesByThingTypeAndKindEqualsAndLabelTypeAndKindEquals(
 								labelSequence.getThingTypeAndKind(), labelSequence.getLabelTypeAndKind())
 						.getSingleResult();
+				logger.info("Found existing LabelSequence: '" + savedLabelSequence.getThingTypeAndKind() + "'' "
```
iamanandkr, line 39:
```suggestion
				logger.info("Found existing LabelSequence: '" + savedLabelSequence.getThingTypeAndKind() + "' '"
```

iamanandkr (approved):
Great description! I don't have much experience writing SQL so read through the documentation [here](https://www.postgresql.org/docs/15/sql-createindex.html) and [here](https://www.postgresql.org/docs/15/indexes-partial.html#INDEXES-PARTIAL-EX3) and it LGTM!

src/main/java/com/labsynch/labseer/service/LabelSequenceServiceImpl.java
brianbolt: Thanks. Ha, I fixed this in the logging above which had the same issue but didn't notice it here. Good eye.

PR #438, opened 2023-04-20
ACAS-654

brianbolt:
## Description
 - BBChem in 2023-2 introduced a new small molecule vs. bio molecule flag to the substructure type fingerprints which is breaking ACAS from storing new fingerprints and also from re-standardizing on upgrades to 2023-2.  
 - This PR truncates the ACAS bbchem tables (which are re-populated on ACAS start) and increases the fingerprint size from `2048` -> to `2112` (64 bit) to account for the additional 1/0 flag.
 
Not in this PR but as a followup we should explore if we can get rid of fingerprint hashing all together with ACAS and just use the substructure service if its feasible.

## Related Issue
ACAS-654

## How Has This Been Tested?
 - Deployed change to QA demo and verified restandardization and substructure search worked.


hweissig (approved):
LGTM!

I checked to see if the size of the substructure column is referred to in more than the two places where this PR changes it to 2112. These appear to be the only ones. The procedure to truncate and alter the tables containing a substructure column looks good.

PR #442, opened 2023-05-16
ACAS-667

brianbolt:
## Description
 - Previous version of this script did not work on fresh instances where hibernate had not yet created the columns. This change installs the columns using flyway.

## Related Issue
ACAS-667

## How Has This Been Tested?
Ran roo code on fresh database and verified the correct columns were created.
Verified upgrading from 2023.1.x, where columns were of the wrong type but exist, resulted in the columns being altered from varchar(255) -> text


bffrost (approved):
Makes sense, sorry I missed this on the original code review!

PR #444, opened 2023-06-14
Acas 1023  allow sel to load experiments with existing case insenitive conservative

philippcheung:


brianbolt (approved):
LGTM.  However, can you fill in the Description in the ticket?

bffrost:
LGTM to me too

PR #445, opened 2023-07-20
ACAS-674: Order corporate name based on similarity match.

iamanandkr:
In the current Compound Registration search form, user has the option to specify "CPMD Number List". ACAS allows users to provide incomplete search terms e.g CMPD-0000004, 4, CMPD-004. It'll also return other results matching the `%X` & `%CMPD-XXX...` pattern. When returning results, we don't apply sorting on the results and it could be the case that CMPD-004 could be filtered out based on the maximum upper limit of results. 

We would here want to apply order, such that results that match the criteria the most are returned first in the results.

## Description

We have two steps in the workflow,
1. Get data sorted from the database.
2. Don't reorder database returned results.

In order to do the first step, we are using [pg_trgm.similarity](https://www.postgresql.org/docs/current/pgtrgm.html#id-1.11.7.44.6) to do a similarity match on the corporate name and user specified search corporate name, and then find the maximum similarity value and then order the results.

We also have other tools to do the similarity match which are stricter, but in the use cases I tried with trigram similarity things worked well. (cloudsql does support [pg_trgm](https://cloud.google.com/sql/docs/postgres/extensions)extension)

In order to do the next step, we are now using an array list instead of a hash set to ensure the order is maintained.

## TODO:
~I have here only updated the similarity logic, but haven't added `pg_trgm` as a dependency and so the search as of now will fail unless you manually install the extension.~

K8s: (@brianbolt 🙏 ) https://crucible.bb.schrodinger.com/cru/SB-21821
Non-K8s: https://github.com/mcneilco/acas-postgres/pull/5

## Related Issue

[ACAS-674](https://jira.schrodinger.com/browse/ACAS-674)

## How Has This Been Tested?

I verified now ordering is consistently getting applied and we have results where the term that matches the best is shown as the first item.

Also added acasclient tests.

I have an updated deployment on: https://anakumar-darwin.k8s.dev.bb.schrodinger.com/acas. Please feel free to experiment any particular use cases.

### DIFF
--- a/src/main/java/com/labsynch/labseer/domain/SaltForm.java
+++ b/src/main/java/com/labsynch/labseer/domain/SaltForm.java
@@ -29,7 +29,9 @@ import javax.persistence.criteria.Join;
 import javax.persistence.criteria.JoinType;
 import javax.persistence.criteria.ParameterExpression;
 import javax.persistence.criteria.Predicate;
+import javax.persistence.criteria.Expression;
 import javax.persistence.criteria.Root;
+import javax.persistence.criteria.Order;
 import javax.validation.constraints.Size;
 
 import com.labsynch.labseer.dto.SearchFormDTO;
@@ -239,6 +241,7 @@ public class SaltForm implements Comparable {
 
 		Predicate[] predicates = new Predicate[0];
 		List<Predicate> predicateList = new ArrayList<Predicate>();
+		List<Order> orderList = new ArrayList<Order>();
 
 		if (searchParams.getChemist() != null && !searchParams.getChemist().equals("anyone")) {
 			logger.debug("incoming chemist :" + searchParams.getChemist().toString());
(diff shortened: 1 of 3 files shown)


brianbolt (changes requested):
This looks really great so far but obviously pending getting pgtrim installed consistently in both K8s and local development stacks.  For the local development stack, we'll need to push a new version of the acas-postgres docker image.  The acas-postgres docker image is what you are using locally and there is a script to create and configure the acas database:
https://github.com/mcneilco/acas-postgres/blob/main/src/initdbs.sh

I'd suggest that you check out this repo, modify it to install the pgtrim extension (on existing and new instances) and build the image like this:
```
docker buildx build --platform linux/arm64/v8,linux/amd64 --tag mcneilco/acas-postgres:release-2023-3.x .
```

You would then need to update your local development stack to point at this image by updating the docker-compose.yaml file to use it.

I'm requesting changes but not to this code, but rather to create a PR for the acas-postgres repository. This image has always been built and pushed manually by myself so when that PR gets approved, I can create a new image and we can update the default docker-compose.yaml to use it.

iamanandkr:
TODO: Update the db image tag once https://github.com/mcneilco/acas-postgres/pull/5 build is available.

iamanandkr (commented):
~TODO: Update the db image tag once https://github.com/mcneilco/acas-postgres/pull/5 build is available.~
https://github.com/mcneilco/acas/pull/1108

brianbolt (approved):
LGTM now that SS-39817 is resolved.

brianbolt:
@iamanandkr Actually, hold off on merging this, I had to revert the install of pg_trgm because of SS-43511 for now.

brianbolt:
Merging this as we worked out SS-43511

PR #453, opened 2023-10-24
ACAS-696: Fix vendorID updates

brianbolt:
## Description
 - Setters and gets with name `vendorID` instead of `vendorId`
 - I believe this was all due to the change in 1.13.7 where we massively updated the roo dependencies (spring and hibernate) and removed roo.  The UI code has always sent and watched for "vendorID" but the roo services were now expecting "vendorId".
 - I also verified that even with this bug, if you saved a lot via the user interface (sending `vendorID` instead of `vendorId`, this did not update the lot.vendorid field.  i.e. The UI was not deleting data.
 
## Related Issue
https://jira.schrodinger.com/browse/ACAS-696

## How Has This Been Tested?
Ran acasclient tests. See new test https://github.com/mcneilco/acasclient/pull/144
Verified manually via the UI that vendorID is properly saved and shown in the UI.

### DIFF
--- a/src/main/java/com/labsynch/labseer/domain/Lot.java
+++ b/src/main/java/com/labsynch/labseer/domain/Lot.java
@@ -148,7 +148,7 @@ public class Lot {
 
     @Size(max = 255)
     @Column(name = "vendorid")
-    private String vendorId;
+    private String vendorID;
 
     @ManyToOne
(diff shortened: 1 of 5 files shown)


bffrost (approved):
LGTM, interesting find

PR #456, opened 2023-12-12
ACAS-738: Fix duplicate parent alias checks dry run

brianbolt:
## Description
  - This fixes multiple issues related to uploading multiple lots to a parent (existing or new) that used a parent alias.
  - 
 
## Issue 1) 2 new lots to an existing parent that has an alias

**To reproduce**
Bulkload a compound with a parent alias
Bulkload 2 lots of the same compound in a file (with or without mapping parent alias)
 
**Expectation**
No error

**Actual outcome**
1 entries had: Within File, Parent Alias XXXXXX is not unique

## Issue 2) 2 new lots with a parent alias

**To reproduce**
Bulkload 2 lots of the same compound in a file (mapping parent alias)
 
**Expectation**
No error

**Actual outcome**
1 entries had: Within File, Parent Alias XXXXXX is not unique

## Related Issue
ACAS-738

## How Has This Been Tested?
Changed acas default to enforce globally unique parent alias (https://github.com/mcneilco/acas/pull/1142) and wrote acasclient tests (https://github.com/mcneilco/acasclient/pull/154)

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/BulkLoadServiceImpl.java
@@ -609,24 +609,7 @@ public class BulkLoadServiceImpl implements BulkLoadService {
 
 		if (validate) {
 			try {
-				parent = validateParentAgainstDryRunCompound(parent, numRecordsRead, results);
-				// Check list of aliases within file being bulkloaded
-				if (!propertiesUtilService.getAllowDuplicateParentAliases()) {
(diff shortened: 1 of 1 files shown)


bffrost (approved):
LGTM. I guess we have several duplicate alias checks in the code, but this is really about shifting when one particular dryRun alias check happens, and making sure it only happens when it's a new parent.

I reviewed your acasclient tests and they look good, so provided those are all passing this seems good to go.

jmcneil86043 (approved):
LGTM

PR #458, opened 2024-04-04
ACAS-710: Code table label text search and short name match query parameters

brianbolt:
## Description
 - Add code table/ddict value label text search which searches for label text and orders by a "best match"
 - Add code table/ddict value short name match which finds the ddict value based on short name in the same route
 - Added maxHits argument to limit results where the default is to not update.
 - Added indexes on label_text to facilitate faster search including tsvector and a lower varcharops which was new to me and handles both lower case seraches and wildcard like search cases:

> The operator classes text_pattern_ops, varchar_pattern_ops, and bpchar_pattern_ops support B-tree indexes on the types text, varchar, and char respectively. The difference from the default operator classes is that the values are compared strictly character by character rather than according to the locale-specific collation rules. This makes these operator classes suitable for use by queries involving pattern matching expressions (LIKE or POSIX regular expressions) when the database does not use the standard “C” locale. As an example, you might index a varchar column like this:

```
CREATE INDEX test_index ON test_table (col varchar_pattern_ops);
```

## Related Issue
ACAS-710

## How Has This Been Tested?
- Ran acasclient tests to verify compatability.
- Uploaded thousands of `data column`/ `column tree` ls_type / ls_kind values to ddict value and verified the code table fields behaved properly in Experiment editor and protocol editor as well as those in each of the endpoint manage fields.
- See related https://github.com/mcneilco/acas/pull/1148 for demo video

### DIFF
--- a/src/main/java/com/labsynch/labseer/domain/DDictValue.java
+++ b/src/main/java/com/labsynch/labseer/domain/DDictValue.java
@@ -599,8 +599,12 @@ public class DDictValue {
         q.setParameter("lsKind", lsKind);
         q.setParameter("labelText", labelText.toLowerCase());
         q.setParameter("formattedLabelText", formattedLabelText);
+        if (maxHits != null) {
+            q = q.setMaxResults(maxHits);
+        }
+
         @SuppressWarnings("unchecked")
-        List<DDictValue> results = q.setMaxResults(maxHits).getResultList();
+        List<DDictValue> results = q.getResultList();
         return results;
     }
 
--- a/src/main/java/com/labsynch/labseer/domain/DDictValue.java
+++ b/src/main/java/com/labsynch/labseer/domain/DDictValue.java
@@ -11,6 +11,7 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.PersistenceContext;
+import javax.persistence.Query;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.TypedQuery;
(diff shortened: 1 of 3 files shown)


src/main/java/com/labsynch/labseer/domain/DDictValue.java
```diff
@@ -568,6 +569,45 @@ public static TypedQuery<DDictValue> findDDictValuesByLsTypeEqualsAndLsKindEqual
...
     }
 
+    public static List<DDictValue> findDDictValuesByLsTypeEqualsAndLsKindEqualsAndLabelTextSearch(String lsType, String lsKind, String labelText, Integer maxHits) {
+        if (lsType == null || lsType.length() == 0)
+            throw new IllegalArgumentException("The lsType argument is required");
+        if (lsKind == null || lsKind.length() == 0)
+            throw new IllegalArgumentException("The lsKind argument is required");
+        EntityManager em = DDictValue.entityManager();
+
+        // Format the search text to match any word in the label text
```
brianbolt, line 579: Below is all the magic.  Split and join all search terms and use to_tsvector serching to match.  I found this to work very well in practice.  The one gotcha is that the ts_rank order by was not perfect.  I wanted to search e.g. "ec50" but it was ranking things like "BLAH-phosphorylated EC50" higher.  Adding a preferrential order by case insenstive exact match made the search function how I wanted.

bffrost (approved):
LGTM. It's interesting to learn about the postgres searching functions. This looks like a reasonable API to allow both exact shortName lookups and text searching of label texts.

PR #461, opened 2024-04-11
ACAS-710: Fix fly way migration numbering conflict

brianbolt:
## Description
 - Fix order to not conflict with `V2.4.2.2__bbchem_small_vs_bio_molecule_fingerprint_bit.sql`
 - 

## Related Issue
ACAS-710

## How Has This Been Tested?
Not tested but ran the `find-sort-flywaymigration.bash` script I wrote last month to avoid this kind of issue but forgot to run here. It finds and sorts flyway migrations in proper version order. It showed the obvious conflict in my last commit.


brianbolt:
Merging this as I broke LD builds.

PR #468, opened 2024-11-22
ACAS-819: Update tomcat to 9.0.97

brianbolt:
## Description
 - acas-roo-server is currently using version tomcat:9.0.62-jre11-openjdk-slim-buster which is subject to [NVD - CVE-2022-29885](https://nvd.nist.gov/vuln/detail/CVE-2022-29885).  Update to tomcat:9.0.97-jdk11-temurin-jammy which is the latest tomcat 9 version https://hub.docker.com/_/tomcat .  Also of note is that the slim-buster image is no longer available but this build works and seems to function properly so I went with it over the others which would have required more docker file build changes.
 - I considered upgrading to latest tomcat [11.0.1](https://github.com/docker-library/tomcat/blob/37acdb35f2684d2c41e3c6ed96cc0403bec02447/11.0/jdk21/temurin-noble/Dockerfile) but decided against it for now
 - Also added a few small fixes to fix legacy warning in docker builds:

```
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 2)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 29)
 ```

## Related Issue
https://schrodinger.atlassian.net/browse/ACAS-819

## How Has This Been Tested?
Built the OSS Indigo version and ran acasclient tests.

### DIFF
--- a/Dockerfile-multistage
+++ b/Dockerfile-multistage
@@ -15,7 +15,7 @@ ADD 	. /src
 RUN 	--mount=type=cache,target=/root/.m2 mvn clean && \
         mvn compile war:war -P ${CHEMISTRY_PACKAGE} 
 
-FROM tomcat:9.0.62-jre11-openjdk-slim-buster
+FROM tomcat:9.0.97-jdk11-temurin-jammy
 
 # Third and Last Step That Requires Significant (Relative) Amount of Time 
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Dockerfile-multistage [hunk 1]
```diff
@@ -15,7 +15,7 @@ ADD 	. /src
 RUN 	--mount=type=cache,target=/root/.m2 mvn clean && \
         mvn compile war:war -P ${CHEMISTRY_PACKAGE} 
 
-FROM tomcat:9.0.62-jre11-openjdk-slim-buster
```
brianbolt, line 18: Main scope of PR is here.

Dockerfile-multistage
```diff
@@ -43,17 +43,17 @@ RUN curl -fsSL https://deb.nodesource.com/setup_$NODE_VERSION | bash - && \
...
   chown -R runner:runner /usr/local/tomcat/
 
 # Allow certificates to be added by runner
-RUN chgrp runner /usr/local/openjdk-11/lib/security/cacerts && \
-    chmod g+w /usr/local/openjdk-11/lib/security/cacerts && \
+RUN chgrp runner $JAVA_HOME/lib/security/cacerts && \
+    chmod g+w $JAVA_HOME/lib/security/cacerts && \
     mkdir -p /usr/lib/jvm/java/jre/lib/security && \
-    ln -s /usr/local/openjdk-11/lib/security/cacerts /usr/lib/jvm/java/jre/lib/security/cacerts
+    ln -s $JAVA_HOME/lib/security/cacerts /usr/lib/jvm/java/jre/lib/security/cacerts
```
brianbolt, line 56: This has to do with where some docker compose files are importing certs into the /usr/lib/jvm/java/jre/lib/security/cacerts locations so leaving that part the same

bffrost (approved):
LGTM. I agree with the decision to stick with Tomcat 9, which has an estimated EOL of 2027 so we've got time.

PR #471, opened 2025-01-16
ACAS-831: Remove extra bit for bio/small molecule flag

brianbolt:
## Description
 - Set substructure fields to be 2048 instead of 2112 (opposite of https://github.com/mcneilco/acas-roo-server/pull/438/files)
 
## Related Issue
ACAS-831

## How Has This Been Tested?
 - Started a 25-1 instance with the broken (larger bit column) and verified the same error when trying to register a structure or do a structure search (ERROR: bit string length 2048 does not match type bit(2112))
 - Updated to this fixed version
 - Verified the columns were all changed from 2112 to 2048.
 - Registered a structure and verified substructure was also working.
 - I didn't test the upgrade path where there were structures registered prior to the upgrade but I did truncate the bbchem_parent_structure table and restarted acas and verified the finger print was detected as missing and added back.


dalejerikson (approved):
LGTM; Makes sense!

bffrost (approved):
LGTM

PR #473, opened 2025-03-04
ACAS-843: Fix molWeight comparison when no changes are made

bffrost:
## Description

The Salt editor was reporting MW changes when none were present. This was due to how Java handles Doubles, and comparing them with `!=` / `==` doesn't work well. Switched to `Double.compare` and tested.

## How Has This Been Tested?

- Tested editing a salt and not changing structure, and confirmed there's no warning shown
- Tested making a structure change to a salt and confirmed that there was a warning about MW change


dalejerikson (approved):
LGTM :)

PR #479, opened 2025-04-01
ACAS-855: Allow multiple workers to work on dry run standardization

brianbolt:
## Description
 - Allows roo to run dry run using multiple workers by batch processing parent ids until there are none missing in the dry run standardization table.
 - Additionally allows roo to continue if a worker fails mid dry run standardization.

## Related Issue
https://schrodinger.atlassian.net/browse/ACAS-855

## How Has This Been Tested?
Tested on K8s with: `kubectl scale deployment -l app.kubernetes.io/name=acas --replicas=4`
Tested locally in docker-compose by updating roo with `deploy: replicas: 3` and settings pointing at a bbchem instance remotely.

I tested by loading 1K compounds and setting the following ACAS settings:

```
# How many compounds to process per loop of dryRunStandariationBatch
server.acas.standardizationBatchSize=1
# 1 second delay for each worker to check if standardization is running (this allows roo workers to pick up that standardization is running quickly
client.cmpdreg.serverSettings.checkForDryRunStandardizationDelay=1000
```

- With these setting I ran dry run standardization with success each time.
 - The fastest settings are still the default 1K batch size but I didn't feel it was necessary to load thousands of compounds to test this out when really for testing this we are looking to if there are any conflicts when the workers run at the same time.
 - I noticed that occasionally ,despite using FOR UPDATE SKIP LOCKED and trying to run each loop in a new transaction, I would still occasionally get an exception with duplicate parent id constraint violation.  Meaning that 2 processes grabbed the same ID.  I tried hard to debug this but couldn't solve it.  Despite this, it would run again after the delay and start processing ids again.  I only saw this issue in the rare occasion that you set the batch to 1 which means that each loop is process very fast.   https://github.com/mcneilco/acas-roo-server/blob/1ae076d8b787af3a8fa6f03b1b201d923b925c24/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java#L231


src/main/webapp/WEB-INF/spring/task-scheduler-config.xml
```diff
@@ -0,0 +1,11 @@
+<beans xmlns="http://www.springframework.org/schema/beans"
```
brianbolt, line 1: This and the changes in web.xml are about enabling springs Scheduling framework.

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
```diff
@@ -968,4 +968,24 @@ public Integer getMaxAutoLotNumber() {
 		return this.maxAutoLotNumber;
 	}
 
+	public long checkForDryRunStandardizationDelay;
+
+	@Value("${client.cmpdreg.serverSettings.checkForDryRunStandardizationDelay}")
```
brianbolt, line 973: This configuration sets the delay/how often roo checks to see if it should participate in an actively running dry run standardization.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -909,5 +909,44 @@ public static Collection<StandardizationDryRunCompound> fromJsonArrayToStandardi
...
+	public static List<Long> fetchUnlockedParentIds(int limit) {
+		String sql = "SELECT p.id " +
+					"FROM parent p " +
+					"WHERE NOT EXISTS (" +
+					"    SELECT 1 " +
+					"    FROM standardization_dry_run_compound s " +
+					"    WHERE s.parent_id = p.id" +
+					") " +
+					"ORDER BY p.id ASC " +
+					"FOR UPDATE SKIP LOCKED " +
```
brianbolt, line 922: This uses FOR UPDATE SKIP LOCKED query.   This allows us to checkout chunks of parent ids which have not yet been added to the standardization dry run compound without getting blocked by other processes.  If you run this query in 2 separate transactions, the first query will select e.g. ids 1-1000 and the next query will select ids 1001-2000.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -909,5 +909,44 @@ public static Collection<StandardizationDryRunCompound> fromJsonArrayToStandardi
...
+				.getResultList();
+
+		// Convert BigInteger to Long
+		return result.stream()
+					.map(BigInteger::longValue)
+					.collect(Collectors.toList());
+	}
+
+	@Transactional
+	public static int countRemainingParentsNotInStandardizationDryRunCompound() {
```
brianbolt, line 938: This second query is important because it's executed to determine how many rows are left to be added to the table.  It's possible that e.g. the final batch of checked out parent ids fail for some reason.  If thats the case, we need a final query to verify that all the ids are in the table before existing the dry run process.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -207,70 +203,51 @@ private HashMap<String, Integer> saveDryRunStructures(HashMap<String, CmpdRegMol
...
-			percent = (float) Math.floor(p * 100f / totalCount);
-			if (percent != previousPercent) {
-				Long currentTime = new Date().getTime();
-				// Output if different from the last time.
-				logger.info("populating standardization dry run table " + percent + "% complete (" + p + " of "
-						+ totalCount + ") average speed (rows/min):"
-						+ (p / ((currentTime - startTime) / 60.0 / 1000.0)));
-				currentTime = new Date().getTime();
-				logger.debug("Time Elapsed:" + (currentTime - startTime));
+		while (true) {
```
brianbolt, line 208: Instead of selecting all Parent ids we now run a while true loop doing chunks of N parent ids until there are non left to process and all ids are in the dry run standardization table,  if there are 0 returned, then it needs to do a final check to make sure all ids are processed, or wait a few seconds and try again.  


Besides freeing up memory so that not even parent id has to be held in memory, this also allows us to continue the process if the processing fails for some reason.  If that happens a new roo worker will come up, check if standardization dry run is "running" and will start populating the dry run table until all ids are present.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -335,11 +312,11 @@ private void dryrunStandardizeBatch(List<Long> pIdGroup, int nonMatchingCmpds, i
 				stndznCompound.setNewMolWeight(null);
 			}
 
-			if (newMolWeight == null || stndznCompound.getParent().getMolWeight() == null) {
+			if (newMolWeight == null || parent.getMolWeight() == null) {
```
brianbolt, line 315: I was noticing this was creating a new transaction every time it was fetching parent then mol weight so I used the already in memory parent instead of having to fetch it fresh every time.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -728,36 +704,29 @@ private StandardizationHistory setCurrentStandardizationDryRunStatus(String stat
 
 	@Override
 	public void executeDryRun() throws CmpdRegMolFormatException, IOException, StandardizerException {
-
-		StandardizationHistory stndznHistory = this.setCurrentStandardizationDryRunStatus("running");
+		logger.info("standardization dry run initialized");
```
brianbolt, line 707: This function is what is actually executed by the API route. Its job is to reset the standardization table (truncate it), create a new standardization history and set it to running. and then kick off a `checkForDryRunStandardization` which starts to populate/dupe check the dry run table.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -767,5 +736,52 @@ public StandardizationHistory getMostRecentStandardizationHistory() {
...
+		stndznHistory
+				.setDryRunStandardizationChangesCount(StandardizationDryRunCompound.getReadyStandardizationChangesCount());
+		stndznHistory.merge();
+		logger.info("Set dry run to " + status);
+	};
+		
+
+
+	@Scheduled(fixedDelayString = "${client.cmpdreg.serverSettings.checkForDryRunStandardizationDelay:60000}")
+    public void checkForDryRunStandardization() throws CmpdRegMolFormatException, IOException, StandardizerException {
```
brianbolt, line 763: This is the actual scheduled task which defaults to 60 seconds. 
It checks to see if standardization is already running on this worker and exits if it is or sets it running if it is not.
 It checks the most recent standardization history to see if it's running, and if it is, it kicks off `runDryRun` which helps populate the the table, run deduplication and finalizes the run.

dalejerikson (approved):
LGTM 

The detailed description and comments as well as the logging within the changes really helped me understand this compared to not having those things :)

PR #480, opened 2025-04-03
ACAS-860: Dupe check standardization performance with sql

brianbolt:

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiStandardizationServicesController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiStandardizationServicesController.java
@@ -58,10 +58,8 @@ public class ApiStandardizationServicesController {
 		HttpHeaders headers = new HttpHeaders();
 		headers.add("Content-Type", "application/json; charset=utf-8");
 		logger.info("checking parent structs and saving to dry run table");
-		int numberOfDisplayChanges = 0;
-		numberOfDisplayChanges = standardizationService.populateStandardizationDryRunTable();
-		logger.info("number of compounds with display change: " + numberOfDisplayChanges);
-		return new ResponseEntity<String>(" Compound check done. " + numberOfDisplayChanges, headers, HttpStatus.OK);
+		standardizationService.populateStandardizationDryRunTable();
+		return new ResponseEntity<String>(" Compound check done.", headers, HttpStatus.OK);
 	}
 
 	@RequestMapping(value = "/dryRun", method = RequestMethod.GET, headers = "Accept=application/json")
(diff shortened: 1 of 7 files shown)


src/main/java/com/labsynch/labseer/api/ApiStandardizationServicesController.java
brianbolt, line 1: We were never doing anything with these integers besides returning them in the string response so I removed it as it was harder to get it with plain sql or if, as we have planned to do batch processing of the standardization steps.

src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
```diff
@@ -65,9 +66,9 @@ public class StandardizationDryRunCompound {
 
 	private boolean asDrawnDisplayChange;
 
-	private int newDuplicateCount;
+	private Integer newDuplicateCount;
 
-	private int existingDuplicateCount;
+	private Integer existingDuplicateCount;
```
brianbolt, line 71: The reason these were changed from lower case int/boolean to upper case Integer, Boolean is to allow it to default to null.  Even though we aren't doing batch processing in the bbchem implementation, this becomes important in the non-bbchem case so that we can batch process rows and know which rows have been processed and which have not.

src/main/resources/db/migration/postgres/V2.4.2.8__standardization_dry_run_indexes.sql
```diff
@@ -0,0 +1,10 @@
+-- ACAS-860s
+-- parent id, registration_status, existing_duplicates_count, new_duplicates_count, changed_structure
+CREATE INDEX IF NOT EXISTS stndzn_dry_run_parent_id_idx on standardization_dry_run_compound(parent_id);
+CREATE INDEX IF NOT EXISTS stndzn_dry_run_registration_status_idx on standardization_dry_run_compound(registration_status);
+CREATE INDEX IF NOT EXISTS stndzn_dry_run_existing_duplicate_count_idx on standardization_dry_run_compound(existing_duplicate_count);
+CREATE INDEX IF NOT EXISTS stndzn_dry_run_new_duplicate_count_idx on standardization_dry_run_compound(new_duplicate_count);
+CREATE INDEX IF NOT EXISTS stndzn_dry_run_changed_structure_idx on standardization_dry_run_compound(changed_structure);
+
+-- Create a unique constraint on parent_id
+CREATE UNIQUE INDEX IF NOT EXISTS stndzn_dry_run_parent_id_unique_idx on standardization_dry_run_compound(parent_id) where parent_id is not null;
```
brianbolt, line 10: These indexes are used in the bbchem implementation as well as in the UI for standardization compounds.  I also introduced a parent unique index to avoid any scenario where a parent could be processed twice by standardization.

src/main/java/com/labsynch/labseer/chemclasses/bbchem/ChemStructureServiceBBChemImpl.java
```diff
@@ -1233,4 +1232,134 @@ public StandardizationSettingsConfigCheckResponseDTO checkStandardizerSettings(S
...
+								  "    SELECT " +
+								  "        p1.id AS parent_id, " +
+								  "        COUNT(DISTINCT p2.corp_name) AS duplicate_count, " +
+								  "        STRING_AGG(DISTINCT p2.corp_name, ';') AS duplicate_corp_names " +
+								  "    FROM " +
+								  "        parent p1 " +
+								  "    JOIN " +
+								  "        standardization_dry_run_compound sdr1 ON p1.id = sdr1.parent_id " +
+								  "    JOIN " +
+								  "        bbchem_parent_structure bps1 ON p1.cd_id = bps1.id " +
```
bffrost, line 1300: Is this join to `bbchem_parent_structure` needed? It doesn't look like we use its data anywhere else in the query

bffrost (approved):
Your SQL looks good (aside from one tiny nit) and this change makes sense. Thanks for all the testing notes. LGTM

dalejerikson (approved):
LGTM!

src/main/java/com/labsynch/labseer/chemclasses/bbchem/ChemStructureServiceBBChemImpl.java
brianbolt: Thanks, I missed that. Fixed.  Had one large query originally and decided to split it and missed the extra join.

PR #477, opened 2025-03-27
ACAS-853: Renaming Salt Abbrev where Salt Abbrev is Appended to Lot Corp Name results in loss of reference to experiment/assay data 

dalejerikson:
Ticket: [Renaming Salt Abbrev where Salt Abbrev is Appended to Lot Corp Name results in loss of reference to experiment/assay data](https://schrodinger.atlassian.net/browse/ACAS-853)

## Description

**Reproduction Steps**

1. For the given server the following should be set `ACAS_CLIENT_CMPDREG_SERVERSETTINGS_APPENDSALTCODETOLOTNAME: "true"`. Therefore the corp batch format should be `corp_batch_saltcode` 
2. Create a given salt 
3. Register compound with salt created in previous step (i.e. The lot corp name should look like `<Prefix>-000XXXX-<SaltAbbrev>`)
4. Register assay data using the lot corp name  
5. Go to the salt browser and edit the salt 
6. Only edit and change the Salt Abbrev 
7. There should be warnings like the following 

`The abbreviation of this salt will be changed`
`This salt is referenced by <x> lots. `
`<x> Associated Batch Code(s): [...]`
`<y> Associated Experiment(s): [...]`
`<z> Associated Container(s): [...] `
`Lot corp names references salt abbreviations. Any associated lot corp names will be updated.`

8. Click to update the salt 

**Previous Bug**

The lot corp name is updated. However, the references to experiments, containers, etc. (like those mentioned above) will be lost. 

**After Fix**

The lot corp names are updated as well as the batch codes referencing the downstream data.

See testing notes below. 

## How Has This Been Tested?

Followed above reproduction steps 

**Looking at Database...**

```
acas=> SELECT id, ignored, code_kind, code_origin, code_type, code_type_and_kind, code_value, ls_kind, modified_by, modified_date, numeric_value, public_data, recorded_by, recorded_date FROM analysis_group_value;
 id | ignored | code_kind  | code_origin  | code_type | code_type_and_kind  |            code_value            |    ls_kind     | modified_by |      modified_date      |     numeric_value     | public_data | recorded_by |      recorded_date
----+---------+------------+--------------+-----------+---------------------+----------------------------------+----------------+-------------+-------------------------+-----------------------+-------------+-------------+-------------------------
 42 | f       |            |              |           | null_null           |                                  | In vitro Clint |             |                         | 12.100000000000000000 | t           | bob         | 2025-03-27 14:57:43.803
[...]

PR #482, opened 2025-04-07
ACAS-856: Changing salts on lots with append salt code to corp name disassociates assay data 

dalejerikson:
## Description

**Previous Bug:** 

Changing the salt on a given lot with append salt code to corp name setting causes dissociation of assay data. The lot corp name is updated. However, the references to experiments, containers, etc. are lost. 

**Steps to Reproduce:** 

- For the given server the following should be set `ACAS_CLIENT_CMPDREG_SERVERSETTINGS_APPENDSALTCODETOLOTNAME: "true"` 
- Therefore the corp batch format should be `corp_batch_saltcode` 
- Create two salts 
- Register compound with one of the salts 
- Register assay data using the lot corp name  
- Go to the lot details page 
- Change the salt to the 2nd salt you created 
- Update the lot which renames the lot to  <Prefix>-000XXXX-NewSaltAbbrev

**Fix Results:** 

- Data associated with lot is modified to the new Lot Corp Name and relationship to assay data is preserved 

## Jira Ticket 
[ACAS-856](https://schrodinger.atlassian.net/browse/ACAS-856 )

## Related Issue
https://github.com/mcneilco/acas-roo-server/pull/477 

## How Has This Been Tested?

Followed reproduction steps prior to bug fix and confirmed issue on local checkout of `2025.2.x` 

Made corresponding changes to `SaltFormServiceImpl.java`. Setup debugger to follow the flow of the `SaltForm` update.

Received confirmation of lot corp name being updated successfully. 

Salt A none

Because this is a similar issue to https://github.com/mcneilco/acas-roo-server/pull/477 , I did a quick check of the database and saw the batch code was updated successfully in one of the assay data tables. 

`select code_value, ignored, deleted, recorded_by, recorded_date  from analysis_group_value agv where ignored = false and code_value is not null;` 

Screenshot 2025-04-07 at 4 36 18 PM

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/SaltFormServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/SaltFormServiceImpl.java
@@ -250,8 +250,10 @@ public class SaltFormServiceImpl implements SaltFormService {
 					// numbering
 					lot.setCorpName(lotService.generateCorpName(lot));
 					newLotCorpName = lot.getCorpName();
-					assayService.renameBatchCode(oldLotCorpName, newLotCorpName, "SaltFormService");
-					containerService.renameBatchCode(oldLotCorpName, newLotCorpName, "SaltFormService", null);
+					if (!oldLotCorpName.equals(newLotCorpName)) {
+						assayService.renameBatchCode(oldLotCorpName, newLotCorpName, "SaltFormService");
+						containerService.renameBatchCode(oldLotCorpName, newLotCorpName, "SaltFormService", null);
+					}
 				}
 				lot.merge();
 				if (!oldLotCorpName.equals(newLotCorpName))
--- a/src/main/java/com/labsynch/labseer/service/SaltFormServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/SaltFormServiceImpl.java
@@ -27,7 +27,13 @@ import org.springframework.stereotype.Service;
 public class SaltFormServiceImpl implements SaltFormService {
 
 	@Autowired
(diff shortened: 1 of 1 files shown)


src/main/java/com/labsynch/labseer/service/SaltFormServiceImpl.java
```diff
@@ -244,6 +250,8 @@ public SaltForm updateSaltForm(Parent parent, SaltForm saltForm, Set<IsoSalt> is
 					// numbering
 					lot.setCorpName(lotService.generateCorpName(lot));
 					newLotCorpName = lot.getCorpName();
+					assayService.renameBatchCode(oldLotCorpName, newLotCorpName, "SaltFormService");
```
brianbolt, line 253: Do you know if this gets hit in the case that a lot hasn't changed names?  i.e. Do we need to test for equality and only call assay/container rename service if the lot corp name has changed?

dalejerikson: Ah thanks for pointing that out. I saw this was within the context of 

```
if (!propertiesUtilService.getCorpBatchFormat().equalsIgnoreCase("cas_style_format")) {
					// if NOT cas_style lot corpName; generate a new lot corp name based on lot
					// numbering
					lot.setCorpName(lotService.generateCorpName(lot));
```

but forgot there are common circumstances where the lot corp name doesn't change so I added a 

`if (!oldLotCorpName.equals(newLotCorpName)) {`

check.

brianbolt (approved):
LGTM!

PR #490, opened 2025-06-27
ACAS-882: Replace p_api_analysis_group_results with fixes for agv2.ignored

bffrost:
## Description

I encountered an issue when the `batch code` AGV is ignored, this API view will still display data against it. This was originally observed in LiveDesign and will be fixed there via DI mappings, but I thought it appropriate to also fix that same bug here.

## How Has This Been Tested?

I tested these changes indirectly by making the same change to LiveDesign Data Integrator mappings and confirmed they fixed the issue.


brianbolt (approved):
LGTM

PR #512, opened 2026-04-01
ACAS-951: Automatic startup restandardization with cluster coordination

brianbolt:
## Summary

Implements ACAS-951 startup auto-restandardization with Kubernetes-safe coordination and rollout hardening, plus auto dry-run report retrieval support.

## Companion PR

This server-side implementation is paired with the ACAS UI/config PR:
- https://github.com/mcneilco/acas/pull/1243

## What Changed

- Adds startup auto-restandardization orchestration in StandardizationServiceImpl.
- Coordinates startup initiation across pods with PostgreSQL advisory locking.
- Adds worker settings-hash guard so mixed-generation pods skip incompatible dry-run work.
- Adds readiness gating so workers wait for standardizer actions initialization before attempting hash-match participation.
- Adds duplicate reservation race tolerance during dry-run population.
- Adds auto dry-run report support:
  - deterministic report naming by history id,
  - endpoint to download report by history id,
  - report-availability flag on history rows (used by ACAS UI to avoid dead links).

### Environment setup used to reproduce and verify mixed rollout behavior

- Forced slower processing to make worker behavior visible:
  - client.cmpdreg.serverSettings.checkForDryRunStandardizationDelay=1000
  - externalStructureProcessingBatchSize: 1
  - standardizationBatchSize: 1
- Verified initial config consistency across running pods:
  - ConfigMap hash scheme: ALL_LAYERS
  - Pod-local conf.properties hash scheme: ALL_LAYERS
- Tailed live pod logs through rolling replacement while changing hash scheme to TAUTOMER_INSENSITIVE_LAYERS.

### Observed rollout sequence

- After ConfigMap update, old and new generations briefly coexisted as expected.
- New pods came online with TAUTOMER_INSENSITIVE_LAYERS while old pods still had ALL_LAYERS.
- Log stream showed expected pod churn: old generation terminating, new generation scaling up.

### What I tested (latest rerun)

- Confirmed startup lock coordination still works:
  - Pod acas-59d6c946fd-kn227 acquired startup cluster lock at 18:22:27.
- Confirmed dry-run history initialization captured expected settings snapshot:
  - historyId=18, settingsHash=810646643, hash_scheme=TAUTOMER_INSENSITIVE_LAYERS.
- Confirmed old-generation pods were correctly excluded:
  - Pod acas-7c855d8599-l7b7j repeatedly logged config mismatch for historyId=18.
  - Mismatch was expected: history hash 810646643 vs pod hash 2081901503 (ALL_LAYERS).
- Confirmed readiness check behavior:
  - New pods first logged readiness wait (standardizer actions are not initialized yet), then joined normally.
  - Examples:
    - acas-59d6c946fd-p7z8g: waits at 18:23:01 and 18:23:02, match at 18:23:03.
    - acas-59d6c946fd-cbvv5: waits at 18:23:47 and 18:23:48, match at 18:23:49.
- Confirmed multi-pod participation and race tolerance:
  - Multiple workers processed batches.
  - Duplicate reservation warnings occurred but processing continued without abort.
- Confirmed end-to-end completion:
  - Dry run completed at 18:23:53.
  - Execute phase started immediately.
  - Auto-restandardization completed successfully at 18:24:41.

## Conclusion

- Readiness gating fixed transient startup false negatives: pods now wait and then join, instead of being misclassified while initializing.
- Settings-hash guard continues to block truly incompatible old-generation workers.
- Mixed rollout behavior is now robust and completes successfully end-to-end.

## How startup cluster lock works

1. A pod opens a dedicated DB connection.
2. It attempts non-blocking advisory lock acquire using string-derived lock ids.
3. If lock is not acquired, that pod exits startup initiation.
4. Lock holder runs: fail stale runs, dry run, verify completion, execute.
5. In finally: advisory unlock attempt plus connection close (session-scoped safety).

This ensures a single startup coordinator pod while preserving distributed dry-run worker processing.

## Detailed testing notes

Building:
acas
```
export DOCKER_HOST=tcp://containerbuilder.onschrodinger.com:23755 SSH_AUTH_SOCK=~/.ssh/ssh_auth_sock && docker build -t mcneilco/acas-oss:$USER .
```

acas-roo-server:
```
export DOCKER_HOST=tcp://containerbuilder.onschrodinger.com:23755 SSH_AUTH_SOCK=~/.ssh/ssh_auth_sock && docker build --build-arg CHEMISTRY_PACKAGE=bbchem -f Dockerfile-multistage -t mcneilco/acas-roo-server-oss:$USER-bbchem .
```

Added these builds to the acas_custom_schrodinger Dockerfiles for roo and acas
Loaded 1000 compounds with stereo category unknown
Then again same file with another stereo to make 2k compounds.

In overrides.yaml added this to make it take a while to restandardize.
```
acas:
  configs:
    client:
      cmpdreg:
        serverSettings:
          checkForDryRunStandardizationDelay: 1000
    server:
      acas:
        externalStructureProcessingBatchSize: 1
        standardizationBatchSize: 1
```

make init

Loaded 1000 compounds with stereo category unknown
Then again same file with another stereo to make 2k compounds.

Requires kubectl krew (plugins)
```
kubectl krew install stern 
```

Functions to log all pods instead of just one:
```
klogpod () {
 APP=$1 
 local CONTAINER=$2 
 local context=$(kubectl config current-context) 
 local PROJECT_ID=$(echo $context | cut -d'_' -f2) 
 local REGION=$(echo $context | cut -d'_' -f3) 
 local CLUSTER=$(echo $context | cut -d'_' -f4-) 
 local QUERY="resource.type=\"k8s_container\" AND resource.labels.cluster_name=\"${CLUSTER}\" AND resource.labels.pod_name=~\"^${APP}\"" 
 if [[ -n "$CONTAINER" ]]
 then
  QUERY="$QUERY AND resource.labels.container_name=\"${CONTAINER}\"" 
 fi
 echo "tailing with query: $QUERY"
 gcloud beta logging tail "$QUERY" --project="$PROJECT_ID" --format='value(timestamp,resource.labels.pod_name,resource.labels.container_name,textPayload)'
}
klogacas () {
 klogpod acas
}
```

With these pods runnings:

```
echo "== configmap =="
kubectl get cm acas-env -o jsonpath='{.data.ACAS_CLIENT_CMPDREG_SERVERSETTINGS_LIVEDESIGN_PREPROCESSORSETTINGS}' | jq '.process_options.hash_scheme'
for pod in $(kubectl get pods -l app.kubernetes.io/name=acas -o jsonpath='{.items[*].metadata.name}'); do
  echo "=== $pod ==="
  kubectl exec "$pod" -c acas -- cat conf/compiled/conf.properties \
    | grep hash_scheme \
    | sed 's/^[^=]*=//' \
    | jq '.process_options.hash_scheme'
done
== configmap ==
"ALL_LAYERS"
=== acas-7c855d8599-l7b7j ===
"ALL_LAYERS"
=== acas-7c855d8599-wb8f6 ===
"ALL_LAYERS"

```

And tailing logs like this (which outputs to stderr the pods starting and stopping.

```
klogacas roo > /tmp/output.txt
+ acas-7c855d8599-l7b7j › roo
+ acas-7c855d8599-wb8f6 › roo
```

Flipped the hash scheme back and forth and watched as things restandardized (had to save values.yaml a few times to get things rolling:
https://github.com/schrodinger/acas_custom_schrodinger/blob/fe1fd0e183b0c078715a44b28f35eaecf9965330/helm/values.yaml#L235

Update the config map to a different has scheme (TAUTOMER_INSENSITIVE_LAYERS)

config map changes take effect:

```
== configmap ==
"TAUTOMER_INSENSITIVE_LAYERS"
=== acas-59d6c946fd-kn227 ===
"TAUTOMER_INSENSITIVE_LAYERS"
=== acas-7c855d8599-l7b7j ===
"ALL_LAYERS"
```

logging also shows new pod generation coming on line and adding additional workers:
```
klogacas roo > /tmp/output.txt
+ acas-7c855d8599-l7b7j › roo
+ acas-7c855d8599-wb8f6 › roo
- acas-7c855d8599-wb8f6 › roo
+ acas-59d6c946fd-kn227 › roo
- acas-7c855d8599-l7b7j › roo
+ acas-59d6c946fd-p7z8g › roo
+ acas-59d6c946fd-cbvv5 › roo
+ acas-59d6c946fd-9jf82 › roo
```

Output analyzed by claude: [output.txt](https://github.com/user-attachments/files/26473289/output.txt)

```
I finished parsing the filtered run output and extracted the latest rollout sequence. Next I’ll give you a ready-to-paste What I’ve tested section with concrete evidence and a short conclusion on what this rerun proves.What I’ve tested (latest rerun)
	•	Verified startup lock behavior:
	◦	One pod acquired the startup cluster lock and initiated auto-restandardization.
	◦	Other pods logged that startup initiation was skipped because lock was already held.
	•	Verified dry-run history initialization captured full settings:
	◦	New dry run history (historyId 17) was initialized with hash 2081901503 and hash_scheme ALL_LAYERS.
	•	Verified config-mismatch guard during mixed rollout generations:
	◦	Old generation pods (5c549b49f8-*) repeatedly skipped dry-run processing due to hash mismatch against historyId 17.
	◦	Their current settings showed TAUTOMER_INSENSITIVE_LAYERS while history expected ALL_LAYERS.
	◦	This confirms old-config pods were blocked from participating.
	•	Verified transient startup instability still occurs on new pods:
	◦	Newly started same-generation pods initially reported mismatch with current settings showing standardizer_actions null.
	◦	After initialization completed, those same pods later logged settings match and started processing dry-run work.
	◦	This reproduces the timing window where effective settings are temporarily incomplete.
	•	Verified multi-pod participation after stabilization:
	◦	Multiple new-generation pods eventually logged Dry-run worker settings match and executed dry-run processing.
	•	Verified duplicate reservation race is now tolerated:
	◦	Many duplicate parent reservation warnings occurred across workers.
	◦	Dry run still progressed to completion instead of failing, confirming retry behavior is active.
	•	Verified full end-to-end completion:
	◦	Dry run transitioned to complete.
	◦	Auto execution started, restandardization ran, and completed successfully.
	◦	Final success log for auto-restandardization was present.
Conclusion
	•	The guard is working as intended for mixed config rollouts.
	•	The duplicate reservation retries prevent run abortion under contention.
	•	The remaining issue is the transient standardizer_actions null window during pod startup, which temporarily blocks otherwise-correct same-generation pods until initialization settles.
If you want, I can draft a second section for your notes called Remaining Risk / Follow-up that proposes one concrete mitigation (readiness gate or delayed scheduler start) in 3-4 bullets.


```

### DIFF
--- a/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
@@ -3,11 +3,16 @@ package com.labsynch.labseer.service;
 import java.io.File;
 import java.io.IOException;
 import java.text.DecimalFormat;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Set;
+import javax.sql.DataSource;
 
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.Query;
@@ -74,7 +79,81 @@ public class StandardizationServiceImpl implements StandardizationService, Appli
 	@Autowired
 	public CmpdRegSDFWriterFactory sdfWriterFactory;
 
+	@Autowired
+	private DataSource dataSource;
+
+	private static final long AUTO_RESTANDARDIZATION_CLUSTER_LOCK_KEY = 951001L;
+
 	Boolean standardizationDryRunRunningInThisWorker = false;
+
+	private boolean tryAcquireClusterLock(Connection connection, long lockKey) throws SQLException {
+		String sql = "SELECT pg_try_advisory_lock(?)";
+		try (PreparedStatement statement = connection.prepareStatement(sql)) {
+			statement.setLong(1, lockKey);
+			try (ResultSet rs = statement.executeQuery()) {
+				if (rs.next()) {
+					return rs.getBoolean(1);
+				}
+			}
+		}
+		return false;
+	}
+
+	private void releaseClusterLock(Connection connection, long lockKey) {
+		String sql = "SELECT pg_advisory_unlock(?)";
+		try (PreparedStatement statement = connection.prepareStatement(sql)) {
+			statement.setLong(1, lockKey);
+			statement.executeQuery();
+		} catch (SQLException e) {
+			logger.warn("Failed to release auto-restandardization cluster lock", e);
+		}
+	}
+
+	private void runAutomaticRestandardizationWithClusterLock() {
+		Connection lockConnection = null;
+		boolean lockAcquired = false;
+		try {
+			lockConnection = dataSource.getConnection();
+			lockAcquired = tryAcquireClusterLock(lockConnection, AUTO_RESTANDARDIZATION_CLUSTER_LOCK_KEY);
+			if (!lockAcquired) {
+				logger.info("Another pod already holds the auto-restandardization startup lock. Skipping startup initiation in this pod.");
+				return;
+			}
+
+			logger.info("Acquired auto-restandardization startup cluster lock.");
+			logger.info("Auto-restandardization: marking stale running standardization records as failed before startup run.");
+			failRunnningStandardization();
+
+			logger.info("Auto-restandardization: starting dry run");
+			executeDryRun();
+
+			StandardizationHistory latestHistory = getMostRecentStandardizationHistory();
+			if (latestHistory == null || !"complete".equals(latestHistory.getDryRunStatus())) {
+				String dryRunStatus = latestHistory == null ? "<missing-history>" : String.valueOf(latestHistory.getDryRunStatus());
+				logger.error("Auto-restandardization: dry run did not complete successfully (status: {}). Skipping execute phase.", dryRunStatus);
+				return;
+			}
+
+			logger.info("Auto-restandardization: dry run complete, starting standardization execution");
+			executeStandardization("acas", "Automatic restandardization triggered on startup");
+			logger.info("Auto-restandardization completed successfully");
+		} catch (SQLException e) {
+			logger.error("Auto-restandardization skipped because cluster lock could not be acquired via database advisory lock", e);
+		} catch (Exception e) {
+			logger.error("Auto-restandardization failed", e);
+		} finally {
+			if (lockAcquired && lockConnection != null) {
+				releaseClusterLock(lockConnection, AUTO_RESTANDARDIZATION_CLUSTER_LOCK_KEY);
+			}
+			if (lockConnection != null) {
+				try {
+					lockConnection.close();
+				} catch (SQLException e) {
+					logger.warn("Failed to close auto-restandardization lock connection", e);
+				}
+			}
+		}
+	}
 
 	public void failRunnningStandardization() {
 		// Cancel all running standardization histories as failed
@@ -126,15 +205,7 @@ public class StandardizationServiceImpl implements StandardizationService, Appli
 			} else {
 				logger.info("Auto-restandardization is enabled and restandardization is needed. Starting automatic restandardization in background thread.");
 				Thread autoRestandardizationThread = new Thread(() -> {
-					try {
-						logger.info("Auto-restandardization: starting dry run");
-						executeDryRun();
-						logger.info("Auto-restandardization: dry run complete, starting standardization execution");
-						executeStandardization("acas", "Automatic restandardization triggered on startup");
-						logger.info("Auto-restandardization completed successfully");
-					} catch (Exception e) {
-						logger.error("Auto-restandardization failed", e);
-					}
+					runAutomaticRestandardizationWithClusterLock();
 				});
 				autoRestandardizationThread.setName("auto-restandardization");
 				autoRestandardizationThread.setDaemon(true);
--- a/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
@@ -322,7 +322,16 @@ public class StandardizationServiceImpl implements StandardizationService, Appli
 	public void populateStandardizationDryRunTable()
 			throws CmpdRegMolFormatException, IOException, StandardizerException {
 		while (true) {
-			int processedCount = dryrunStandardizeBatch();
+			int processedCount;
+			try {
+				processedCount = dryrunStandardizeBatch();
+			} catch (RuntimeException e) {
+				if (isDuplicateDryRunParentReservation(e)) {
+					logger.warn("Detected duplicate parent reservation while populating standardization dry run table. Another worker likely won the race for one or more parent IDs. Retrying.", e);
+					continue;
+				}
+				throw e;
+			}
 			int remainingToBeProcessed = StandardizationDryRunCompound.countRemainingParentsNotInStandardizationDryRunCompound();
 			if(processedCount > 0) {
 				int totalProcessed = StandardizationDryRunCompound.rowCount();
@@ -342,6 +351,20 @@ public class StandardizationServiceImpl implements StandardizationService, Appli
 				}
 			}
 		}
+	}
+
+	private boolean isDuplicateDryRunParentReservation(Throwable throwable) {
+		Throwable current = throwable;
+		while (current != null) {
+			String message = current.getMessage();
+			if (message != null
+					&& message.contains("stndzn_dry_run_parent_id_unique_idx")
+					&& message.contains("duplicate key value violates unique constraint")) {
+				return true;
+			}
+			current = current.getCause();
+		}
+		return false;
 	}
 
 	@Transactional(propagation = Propagation.REQUIRES_NEW)
--- a/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
+++ b/src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
@@ -15,6 +15,8 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.sql.DataSource;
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
```diff
@@ -228,6 +367,95 @@ public void populateStandardizationDryRunTable()
...
+			JsonNode parsedSettings = OBJECT_MAPPER.readTree(standardizerSettings.getSettings());
+			JsonNode standardizerActions = parsedSettings.get("standardizer_actions");
+			return standardizerActions != null && !standardizerActions.isNull() && !standardizerActions.isMissingNode();
+		} catch (IOException e) {
+			logger.warn("Unable to parse standardizer settings while checking readiness. Continuing with normal processing.", e);
+			return true;
+		}
+	}
+
+	private void maybeGenerateAutoRestandardizationDryRunReport(StandardizationHistory latestHistory) {
```
bffrost, line 398: nit: I find the "maybe" in the function name confusing. I guess any function with error handling might not achieve what the caller intends, but I think that's generally understood.

I think `generateAutoRestandardizationDryRunReport` is a fine name for this function.

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
```diff
@@ -988,4 +988,40 @@ public long getCheckForDryRunStandardizationDelay() {
...
+	}
+
+	@Override
+	public Boolean getAutoRestandardize() {
+		return this.autoRestandardize;
+	}
+
+	boolean autoRestandardizationReportEnabled;
+
+	@Value("${client.cmpdreg.serverSettings.autoRestandardizationReportEnabled:false}")
```
bffrost, line 1005: Do we need a separate config for this? Or can we just roll this into autoRestandardize?

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
```diff
@@ -988,4 +988,40 @@ public long getCheckForDryRunStandardizationDelay() {
...
+	}
+
+	@Override
+	public Boolean getAutoRestandardizationReportEnabled() {
+		return this.autoRestandardizationReportEnabled;
+	}
+
+	String autoRestandardizationReportDirectory;
+
+	@Value("${client.cmpdreg.serverSettings.autoRestandardizationReportDirectory:/tmp}")
```
bffrost, line 1017: Should this default value be somewhere in privateUploads so they're retained? I could imagine this history being quite useful

bffrost (commented):
There's a bunch in here but it makes sense to me in general. The startup checks and the locking behavior make sense to me. As I understand, we already store dry run metadata but not actual dry run results beside the "current one" when the server is between dry run and execute, so the new DryRunReport is an important new data store that isn't easily recoverable.

I'd recommend making the reports enabled and stored in privateUploads whenever auto restandardization is done.

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
brianbolt: I was leaving this as a separate config because I wanted to avoid a scenario where the writing of the report file doesn't work well. In that scenario, someone could keep auto restandardizes ON but turn off the report feature because they did this on stage and already confirmed the changes. Basically a FF safety for just the reporting.

That being said I will change this to true by default to match what acas has which is true:

https://github.com/mcneilco/acas/pull/1243/changes#diff-5a0cf16bf04a9805e2d1fe93d86a85c53c15fe1a5bbbec43510385442c29b096R791

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
brianbolt: Yes, ACAS sets this but it does make sense to match the acas default: https://github.com/mcneilco/acas/pull/1243/changes#diff-5a0cf16bf04a9805e2d1fe93d86a85c53c15fe1a5bbbec43510385442c29b096R795

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
brianbolt: Although I do realize after writing this that the reporting feature should maybe not be tied to autoRestandardization at all.  Such that it does report by default whether it's manual or auto.  I'll make that change.

src/main/java/com/labsynch/labseer/service/StandardizationServiceImpl.java
brianbolt: Thanks, made that change

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
brianbolt: I made the above change

src/main/java/com/labsynch/labseer/utils/PropertiesUtilServiceImpl.java
brianbolt: I made this change

brianbolt:
Summary of latest changes:

- Renamed config properties: `autoRestandardizationReportEnabled` → `standardizationDryRunReportEnabled` and `autoRestandardizationReportDirectory` → `standardizationDryRunReportDirectory` (addresses PR feedback)
- Changed default for dry-run report enabled to `true` and updated default directory to `/home/runner/build/privateUploads/cmpdreg/standardizationReports/` (addresses PR feedback)
- Decoupled dry-run report generation from auto-restandardization; now supports both manual and automatic dry-run flows
- Made \`generateDryRunReport\` public and removed unnecessary wrapper method (addresses PR feedback on method naming and visibility)
- Cleaned up imports and interface signatures
- Synced config and code with ACAS repo

brianbolt:
I also retested the basics on a LiveDesign cluster with auto on and off and report files on and off in each mode.

bffrost (approved):
LGTM

PR #516, opened 2026-06-09
ACAS-988: Migrate roo runtime to minimal pinned glibc base (debian:12-slim)

brianbolt:
## Problem
The roo runtime base `tomcat:10.1-jdk21-temurin` is a full Ubuntu image, uses the JDK (not JRE), and is unpinned (floats the patch) — violating the [container image security baselines](https://schrodinger.atlassian.net/wiki/spaces/LDE/pages/1940160531) (minimal, version-pinned runtime). Build-stage bases (`maven:3-eclipse-temurin-21`, `node:20-slim`) were also unpinned.

## Change
- **Runtime → `debian:12-slim`** with the Temurin **JRE** + Tomcat copied from the pinned `tomcat:10.1.55-jre21-temurin` image. `-slim` is baseline-compliant.
- **glibc is required, not Alpine.** Indigo's native chemistry library (`libindigo.so`) is glibc-linked in every published version (verified: imports `GLIBC_2.x`, needs `libc.so.6`/`libgcc_s.so.1`; EPAM publishes no musl/Alpine build). An Alpine attempt deployed but the `/acas` context failed to start because JNA could not load `libindigo.so` on musl. `debian:12-slim` keeps glibc while still being minimal. `libstdc++6`/`libgomp1` are installed for the native chem libs; `node` is sourced from the glibc `node:20.20.2-slim`.
- **Build stages pinned:** `maven:3.9.9-eclipse-temurin-21`, node `20.20.2`.

## Testing
Built with `-P indigo` and run via docker-compose: roo deploys `/acas`, indigo loads, and `GET /acas/api/v1/protocoltypes` returns 200.

also ran acasclient tests

also ran bbchem flavor + acas loader files which do a bulk load against bbchem.
as part of that test ran open in LiveDesign button and also made sure curve rendering worked inline in LiveDesign.

Part of ACAS-988.

### DIFF
--- a/Dockerfile-multistage
+++ b/Dockerfile-multistage
@@ -1,7 +1,11 @@
 ARG 	CHEMISTRY_PACKAGE=bbchem
-ARG     NODE_VERSION=20
-FROM maven:3-eclipse-temurin-21 AS builder
+ARG     NODE_VERSION=20.20.2
+ARG     TOMCAT_VERSION=10.1.55
+FROM maven:3.9.9-eclipse-temurin-21 AS builder
+# glibc node (Debian) so it matches the debian-slim runtime below
 FROM node:${NODE_VERSION}-slim AS node
+# Pinned source of the Temurin JRE + Tomcat distribution (copied into the runtime)
+FROM tomcat:${TOMCAT_VERSION}-jre21-temurin AS tomcat
 
 FROM 	builder AS compile
 ARG     CHEMISTRY_PACKAGE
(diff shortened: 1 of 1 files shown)


bffrost:
@brianbolt have you tested the bbchem flavor of this too? Otherwise this looks reasonable to me: pinning versions and copying from various repos into the minimal runtime images. All comes down to testing.

brianbolt:
@bffrost No, i have not and found a few issues with it.  Testing now but this also requires ACAS-979 (install maestro sketcher without make + gcc) so I'm working on that ticket as well as a blocker now.

brianbolt:
@bffrost I have now tested bbchem and it works, the fixes for the problems I mentioned are in these related PRs:

- acas oss related changes required "truststore" which is new so I included here: https://github.com/mcneilco/acas/pull/1246
- Install rather than build sketcher https://github.com/schrodinger/acas_custom_schrodinger/pull/194

bffrost (approved):
LGTM

PR #517, opened 2026-06-11
ACAS-951: Fix dry-run restandardization deadlocks (stale snapshot + cross-replica DB deadlock)

brianbolt:
Fixes three related defects in the restandardization dry-run flow (`StandardizationServiceImpl`) that surfaced after auto-restandardization-on-startup (#512), in a multi-replica cluster during/after a chemistry suite version bump.

## Bug 1 — a dry run keeps a stale copy of the chemistry settings

Each dry run is tracked by a `standardization_history` row with a `dry_run_status` (`running` → `complete`) and a `settings_hash` (the hash of the chemistry settings it started with). A worker pod only joins a run if its live settings hash equals the row's `settings_hash`. `setCurrentStandardizationDryRunStatus` starts a run by either creating a new row, or **reusing the most-recent row when its `standardization_status` isn't `complete`** — which is common, because a plain `/dryRun` (and `failRunnningStandardization`) leave `standardization_status` NULL. It wrote `settings_hash` only on the create path, so a reused row kept its **old** hash. After a suite upgrade (Build 058 → 073) the reused row still advertised 058; every pod's live hash (073) failed the check and skipped (`config mismatch detected … waiting for pod to restart`, every ~60s), and the row sat at `running` forever.

```
The most-recent standardization_history row is from a prior dry run that never ran a full
standardization, so its standardization_status is still NULL (a plain /dryRun never sets it).

A new dry run starts on a Build 073 pod. Because that row isn't 'complete', the code REUSES it:
it flips dry_run_status back to 'running' but does NOT refresh settings_hash -> stays 058.   <- bug

Worker check before joining:  live settings_hash 073  ==  row.settings_hash 058 ?
   -> no -> "config mismatch ... waiting for pod to restart" -> skip
   ...nothing is on 058 anymore, so it never matches and the row sits at 'running' forever.
```

**Fix:** always write the current `settings_hash` on initiation — new or reused row (pure, unit-tested helper `applyDryRunInitialization`).

## Bug 2 — starting a second dry run while one is running deadlocks the database

Dry-run population is intentionally cooperative — every replica claims disjoint batches via `SELECT … FOR UPDATE SKIP LOCKED`, in parallel. The defect was at the lifecycle edges: nothing stopped a *second* dry run from starting while one was already running. Starting calls `reset()`, which `TRUNCATE`s the dry-run tables; that TRUNCATE collided with the in-flight `INSERT`s (`standardization_dry_run_compound` AccessExclusiveLock vs `bbchem_standardization_dry_run_structure` RowExclusiveLock) → `deadlock detected`, and two pods marking the run `running` clashed on the history row (optimistic-lock). The run was marked `failed`.

```
Intended (kept):  one run "running"
   Pod A - claim batch 1 (SKIP LOCKED) -> INSERT
   Pod B - claim batch 2 (SKIP LOCKED) -> INSERT   <- disjoint rows, in parallel

Bug:  a 2nd dry run STARTS mid-populate
   Pod B - reset(): TRUNCATE standardization_dry_run_compound   (table lock)
   Pod A - still INSERTing bbchem_standardization_dry_run_structure (row locks)  -> deadlock -> "failed"
```

**Fix:** keep population parallel and lock-free; guard only the lifecycle edges with a cluster-wide `pg_try_advisory_lock` (`ADVISORY_LOCK_DRY_RUN_EXECUTION`) — **start** (`reset()` + mark running, only when no run is active) and **finish** (dupe-check + finalize, by whichever pod drains the queue first). A TRUNCATE then never overlaps population.

## Bug 3 — a pod dying mid-run lets the next pod reset on top of live workers

This is the trigger behind the original incident. On a (re)starting pod, auto-restandardization runs `failRunnningStandardization()` — which marked **every** `running` dry run as `failed` — and then `executeDryRun()` (reset + start). If the pod running a dry run was terminated mid-populate while *other* pods were still cooperatively populating that run, the new pod's `failRunnningStandardization` flipped the still-active run to `failed`, so the `reset()` that follows `TRUNCATE`d the dry-run tables while those workers were still `INSERT`ing → the Bug 2 deadlock again, plus re-processed parents. It slips past Bug 2's "only reset if not running" guard because the status was changed out from under live workers.

```
Auto-restandardization on a (re)starting pod runs:  failRunnningStandardization()  then  reset()

Bug:  the pod running the dry run is terminated mid-populate; another pod keeps INSERTing.
   New pod : failRunnningStandardization() flips the still-active run 'running' -> 'failed'
             reset(): TRUNCATE  ──┐ races
   Other pod: still INSERTing  ───┴► deadlock detected (+ parents re-processed)

Fix:  only fail a run with NO progress for > dryRunStaleMillis (default 5 min).
   "progress" = newest qc_date in standardization_dry_run_compound (every batch stamps it).
   An actively-progressing run is left 'running' -> the new pod JOINS it, never resets.
```

**Fix:** `failRunnningStandardization` only fails a `running` dry run if it's **stale** — no new populate batch (`qc_date`) for `dryRunStaleMillis` (default 5 min, configurable). An actively-progressing run is left running so the new pod joins it. (A run abandoned by a dead pod but still valid is simply resumed; only one that genuinely can't progress is failed after the timeout and cleanly re-initiated.)

## Changes
- `StandardizationServiceImpl`:
  - `setCurrentStandardizationDryRunStatus` always snapshots current settings (pure helper `applyDryRunInitialization`).
  - New `ADVISORY_LOCK_DRY_RUN_EXECUTION`. `executeDryRun` → `initiateDryRunIfIdle()` (cluster-locked; `reset()` + mark running only if idle) then `participateInDryRun()`. `@Scheduled checkForDryRunStandardization` → `participateInDryRun()` (cooperative, no lock). `runDryRun` populates lock-free, then `finishDryRunUnderClusterLock()` (cluster-locked dupe-check + finalize).
  - `failRunnningStandardization` fails a dry run only if stale (`isDryRunStale`, threshold `dryRunStaleMillis`, default 5 min).
  - Clarified the config-mismatch log message.
- `StandardizationDryRunCompound.findLatestQcDate()` — `max(qc_date)` progress signal for staleness.
- New pure unit test `StandardizationServiceImplTest` (no Spring/DB).

## Why this was missed previously
- Earlier testing changed `hash_scheme` and ran a restandardization **to completion** — that exercises the *create-a-new-row* path, which always wrote `settings_hash` correctly. The bug lives in the *reuse-an-existing-row* path.
- The reuse path only triggers when the latest `standardization_history` row isn't `complete` (e.g. after a plain `/dryRun`, or an interrupted/auto run). A clean "edit config → restandardize → done" loop never leaves such a row around.
- `schrodinger_suite_version` isn't an editable setting — it's read from the BBChem health endpoint, so it only changes when a **new chemistry image is deployed**, i.e. alongside a pod restart. That restart is exactly what both leaves a reusable not-yet-`complete` row *and* changes the hash (Bug 1), and what kills a pod mid-run (Bug 3).
- The whole class only became reachable after #512 made restarts **auto-initiate** a dry run.

## Testing performed
All on dev cluster `shared-dev-cluster-26-2` (bbolt-darwin), BBChem engine, 2–4 roo replicas, ~5,257 parents.

- **Build:** `StandardizationServiceImplTest` 2/2; `-P indigo` test-compile BUILD SUCCESS (Docker JDK 21).
- **Bug 1:** reproduced on the old image (config-mismatch spam, no recovery); fixed image re-snapshots on reuse → live hash matches → run `complete`, spam stops.
- **Bug 2:** reproduced on the old image (two overlapping `/dryRun` → pod A `deadlock detected`, pod B optimistic-lock, run `failed`); fixed image → both pods populate disjoint batches in parallel, the 2nd initiator skips reset, one pod finalizes, run `complete`, **0** deadlock/optimistic-lock (incl. a 300m-CPU slow run for a wide overlap window).
- **Bug 3:** reproduced on the pre-staleness build (pod terminated mid-run during a rolling restart → the next pod's `failRunnningStandardization` clobbered the active run → reset-vs-populate `deadlock detected` + re-processing). With the staleness guard: invoked `failRunnningStandardization` against a genuinely-active run (record `running`, fresh `qc_date`) → it logged `… still making progress; leaving it running so this pod can join it instead of resetting`, did **not** fail the run, the run completed, and there were **0 deadlocks** across all pods.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/test/java/com/labsynch/labseer/service/StandardizationServiceImplTest.java
+++ b/src/test/java/com/labsynch/labseer/service/StandardizationServiceImplTest.java
@@ -1,0 +1,61 @@
+package com.labsynch.labseer.service;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+
+import java.util.Date;
+
+import com.labsynch.labseer.domain.StandardizationHistory;
+import com.labsynch.labseer.dto.configuration.StandardizerSettingsConfigDTO;
+
+import org.junit.Test;
+
+/**
+ * Pure unit tests (no Spring context / DB) for the dry-run initialization snapshot logic.
+ * Guards against the config-mismatch deadlock seen when the chemistry suite version changes
+ * while an incomplete dry-run record is reused.
+ */
+public class StandardizationServiceImplTest {
+
+	private StandardizerSettingsConfigDTO settingsForSuite(String suiteVersion) {
+		StandardizerSettingsConfigDTO dto = new StandardizerSettingsConfigDTO();
+		dto.setShouldStandardize(true);
+		dto.setType("bbchem");
+		dto.setSettings("{\"standardizer_actions\":{\"NEUTRALIZE\":true},"
+				+ "\"hash_scheme\":\"TAUTOMER_INSENSITIVE_LAYERS\","
+				+ "\"schrodinger_suite_version\":\"" + suiteVersion + "\"}");
+		return dto;
+	}
+
+	@Test
+	public void reusedHistoryAdoptsCurrentSettingsSnapshot() {
+		// An orphaned, incomplete dry-run record snapshotted under an OLD suite build.
+		StandardizerSettingsConfigDTO oldSettings = settingsForSuite("Schrodinger Suite 2026-3, Build 058");
+		StandardizationHistory history = new StandardizationHistory();
+		history.setSettings(oldSettings.toJson());
+		history.setSettingsHash(oldSettings.hashCode());
+		history.setDryRunStatus("failed");
+		history.setDryRunComplete(new Date());
+
+		// The initiating pod now runs with a NEW suite build and re-initializes the dry run.
+		StandardizerSettingsConfigDTO currentSettings = settingsForSuite("Schrodinger Suite 2026-3, Build 073");
+		Date start = new Date();
+		StandardizationServiceImpl.applyDryRunInitialization(history, currentSettings, "running", start);
+
+		// Snapshot must reflect the CURRENT config so the worker guard matches and proceeds.
+		assertEquals(currentSettings.hashCode(), history.getSettingsHash());
+		assertEquals(currentSettings.toJson(), history.getSettings());
+		assertEquals("running", history.getDryRunStatus());
+		assertEquals(start, history.getDryRunStart());
+		assertNull(history.getDryRunComplete());
+	}
+
+	@Test
+	public void suiteVersionChangeChangesTheHash() {
+		// Documents the exact trigger: identical actions, different suite build => different hash.
+		int oldHash = settingsForSuite("Schrodinger Suite 2026-3, Build 058").hashCode();
+		int newHash = settingsForSuite("Schrodinger Suite 2026-3, Build 073").hashCode();
+		assertFalse("suite version must affect the settings hash", oldHash == newHash);
+	}
+}
--- a/src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
+++ b/src/main/java/com/labsynch/labseer/domain/StandardizationDryRunCompound.java
@@ -151,6 +151,16 @@ public class StandardizationDryRunCompound {
 				.intValue();
 	}
 
+	/**
+	 * Latest qc_date across all dry-run compounds, i.e. when the most recent populate batch was
+	 * written. Used as a liveness/progress signal for the active dry run (null if the table is empty).
+	 */
+	public static Date findLatestQcDate() {
+		return StandardizationDryRunCompound.entityManager()
+				.createQuery("SELECT max(o.qcDate) FROM StandardizationDryRunCompound o", Date.class)
+				.getSingleResult();
+	}
+
 	public StandardizationHistory fetchStats() {
 		StandardizationHistory stats = new StandardizationHistory();
 		stats.setStructuresStandardizedCount(toIntExact(StandardizationDryRunCompound.entityManager()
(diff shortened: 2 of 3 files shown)


src/test/java/com/labsynch/labseer/service/StandardizationServiceImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	@Test
+	public void reusedHistoryAdoptsCurrentSettingsSnapshot() {
+		// An orphaned, incomplete dry-run record snapshotted under an OLD suite build.
+		StandardizerSettingsConfigDTO oldSettings = settingsForSuite("Schrodinger Suite 2026-3, Build 058");
+		StandardizationHistory history = new StandardizationHistory();
+		history.setSettings(oldSettings.toJson());
+		history.setSettingsHash(oldSettings.hashCode());
+		history.setDryRunStatus("failed");
```
bffrost, line 38: Trying to follow along, should this status be null rather than "failed" to test the bug mentioned in the PR description?

src/test/java/com/labsynch/labseer/service/StandardizationServiceImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+		history.setDryRunComplete(new Date());
```
bffrost, line 39: similarly, should this completion date be null if this is an "orphaned, incomplete dry-run record"?

src/test/java/com/labsynch/labseer/service/StandardizationServiceImplTest.java
bffrost: I think I answered my own question during the review. This reflects the state after `failRunnningStandardization` has been run and marked this as failed.

bffrost (approved):
LGTM. The bug explanations are clear enough and these fixes - failing stale runs, filling in the standardization settings at initialization, and cluster-wide locking during init and finalization make sense to me.

We'll need to see how this behaves on QA systems to be sure about the fixes.

PR #518, opened 2026-06-15
ACAS-923: Return 404 for missing protocol experiments

bffrost:
## Description

A while ago I noticed a quirk in our API where we were returning 500 errors when trying to look up non-existent protocols. This got in the way of my troubleshooting of a legitimate issue (OOM errors caused by protocols with many experiments) so I'm trying to clean it up. The fix is over 3 PRs:
- This PR to update the `throws` and ensure Roo responds with a 404 when the protocol code doesn't exist
- https://github.com/mcneilco/acas/pull/1248 Node PR to bypass "throw 500 on any non-200 response" rule built into node layer
- https://github.com/mcneilco/acasclient/pull/189 acasclient test & removal of workaround that was converting 500s to `None`

## How Has This Been Tested?

On local docker compose stack:
- http://localhost:8080/acas/api/v1/experiments/protocol/FAKECODE returns 404 rather than 500

### DIFF
--- a/src/main/java/com/labsynch/labseer/api/ApiExperimentController.java
+++ b/src/main/java/com/labsynch/labseer/api/ApiExperimentController.java
@@ -1644,7 +1644,7 @@ public class ApiExperimentController {
 			return new ResponseEntity<String>(
 					Experiment.toJsonArrayStub(experiments),
 					headers, HttpStatus.OK);
-		} catch (NoResultException e) {
+		} catch (NotFoundException e) {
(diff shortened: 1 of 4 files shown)


brianbolt (approved):
LGTM

But I think there is a built in check for this in acasclient here: https://github.com/mcneilco/acasclient/blob/90533851f0919170172c517efe2be2814481bae7/acasclient/acasclient.py#L1120

WIth this change, this get_experiment_by_protocol_code acasclient function would start raising instead of returning None as raise for status errors on 404.  we should probably just change that line to:

```
if resp.status_code == 500 or resp.status_code == 404:
```

